PR #131, opened 2026-04-22
feat: Context Contributor pipeline (#107)

forinda:
Closes #107.

Lands the typed pre-handler ctx-extension primitive proposed in the tracking issue. Eight phases, 19 commits, all green via the pre-commit hook (build + test + typecheck + format:check) on every commit.

## What's now usable

```ts
import {
  bootstrap,
  defineContextDecorator,
  Controller,
  Get,
  type RequestContext,
} from '@forinda/kickjs'

declare module '@forinda/kickjs' {
  interface ContextMeta {
    tenant: { id: string }
    project: { id: string; tenantId: string }
  }
}

const LoadTenant = defineContextDecorator({
  key: 'tenant',
  resolve: (ctx) => ({ id: ctx.req.headers['x-tenant-id'] as string }),
})

const LoadProject = defineContextDecorator({
  key: 'project',
  dependsOn: ['tenant'],                 // topo-sorted at boot; cycles fail setup()
  deps: { repo: PROJECT_REPO },          // resolved through DI
  resolve: (ctx, { repo }) => repo.find(ctx.get('tenant')!.id, ctx.params.id as string),
})

@LoadTenant
@Controller('/projects')
class ProjectsController {
  @LoadProject
  @Get('/:id')
  getOne(ctx: RequestContext) {
    return ctx.json({ tenant: ctx.get('tenant'), project: ctx.get('project') })
  }
}

bootstrap({
  modules,
  adapters: [new TenantAdapter()],       // adapter.contributors?()
  contributors: [StartedAt.registration], // ApplicationOptions.contributors
})
```

Five registration sites, precedence high → low: **method > class > module > adapter > global**. Same-precedence collisions throw `DuplicateContributorError`; cross-level collisions are silent overrides. Cycles + missing `dependsOn` keys throw at `app.setup()`, never per request.

## Phases

| Phase | Commits | Highlights |
|---|---|---|
| Architecture | 1 | `architecture.md` §20 — design + locked decisions |
| 1 — Core types | 6 | `defineContextDecorator`, `ExecutionContext`, three error classes, root re-exports |
| 2 — Pipeline | 2 | Builder + Kahn topo-sort + sequential async runner with §20.9 error matrix |
| 3 — Storage unification | 1 | `RequestContext.metadata` routes through `requestStore.values`; `req.__ctxMeta` deprecated fallback |
[...]

PR #132, opened 2026-04-22
feat(auth): type-narrow @Roles and @Can via augmentation

forinda:
Type-narrow `@Roles` and `@Can` (and the matching `AuthorizationService` runtime methods) via two augmentable interfaces. Adopters catch role/policy typos at the decoration site; unaugmented apps see zero behavior change.

## Why

`@Roles('admni')` and `@Can('typo', 'post')` currently sit silently in code until a request hits them — by then it's a 500 or a denied request, not a build failure. Both the decorator metadata layer and the `AuthorizationService` already had string-typed entrypoints; this PR threads a generic over both.

### `AuthUser['roles']` (existing interface)

Adopters already augment this for `ctx.user` typing:

```ts
declare module '@forinda/kickjs-auth' {
  interface AuthUser {
    id: string
    roles: ('admin' | 'editor' | 'viewer')[]
  }
}

@Roles('admin', 'editor')   // ✓
@Roles('typo')              // ✗ TS error
```

### `PolicyRegistry` (new, empty interface)

Map of (resource → declared actions). Augment to narrow `@Can` and `AuthorizationService.{can,listObjects}`:

```ts
declare module '@forinda/kickjs-auth' {
  interface PolicyRegistry {
    post: 'create' | 'update' | 'delete' | 'publish'
    user: 'invite' | 'suspend'
  }
}

@Can('delete', 'post')      // ✓
@Can('typo', 'post')        // ✗ TS error: 'typo' not in 'post' actions
@Can('delete', 'unknown')   // ✗ TS error: 'unknown' not a registered resource
@Can('invite', 'post')      // ✗ TS error: 'invite' is on 'user', not 'post'

await authz.can(user, 'delete', 'post')           // ✓
await authz.listObjects(user, 'view', 'post')     // ✓
```

## Type definitions

```ts
type Role = AuthUser['roles'] extends readonly (infer T)[] ? T : string

type PolicyResource = keyof PolicyRegistry extends never
  ? string
  : keyof PolicyRegistry & string

type PolicyAction<R> = R extends keyof PolicyRegistry
  ? PolicyRegistry[R] extends string ? PolicyRegistry[R] : string
  : string
```

The `extends never` pivot is what gives backwards compatibility — when `PolicyRegistry` is empty (default), `keyof` yields `never` and both `PolicyResource` and `PolicyAction<R>` collapse to `string`. Augmented apps get the literal-union narrowing automatically.

## Runtime impact: zero

- `PolicyRegistry` is an empty interface with **no runtime presence**.
- `policyRegistry: Map<string, any>` (the runtime one) is unchanged.
- `@Policy('post')` registers exactly as before.
- `AuthorizationService.can()` / `listObjects()` bodies untouched — only their signatures gained a generic.
- 150 existing auth tests (which do NOT augment either interface) all pass on the loose-string fallback path, proving back-compat empirically.

## Tests

- 9 new tests in `packages/auth/__tests__/typed-decorators.test.ts` cover both runtime metadata storage AND compile-time rejection via `@ts-expect-error` guards. Those guards are load-bearing — if narrowing regresses, TS no longer sees an error to suppress and the build fails. They functioned as designed (verified locally by removing one and watching it fail).

## Test plan

- [x] `pnpm test` — 159/159 auth tests pass (was 150, +9 new); other packages green
- [x] `pnpm typecheck` — green across all 23 packages
- [x] `pnpm format:check` — clean
- [x] `pnpm docs:build` — clean (210s)
- [x] `@Policy('post')` runtime registration verified via existing policy tests (still 13/13 in `decorators.test.ts`)

## Related work

- Branched from clean `main`, **not** from `feat/context-contributors` (#131). Truly orthogonal — this PR can land independently.
- Out of scope: narrowing `AuthorizationServiceOptions.allow`/`deny` (those accept both `'resource'` and `'resource.action'` string forms; narrowing the union is awkward) and narrowing the user-supplied `listObjects` callback in the constructor (would force adopters to write generic implementations).

### DIFF
--- a/docs/guide/authorization.md
+++ b/docs/guide/authorization.md
@@ -154,11 +154,11 @@ declare module '@forinda/kickjs-auth' {
 @Can('invite', 'post')      // ✗ TS error: 'invite' is not assignable to actions of 'post' (it's on 'user')
 ```
 
-The same narrowing applies to `AuthorizationService.can()` and `AuthorizationService.listObjects()` — pass `<R extends PolicyResource>` to get type-safe runtime checks:
+The same `PolicyRegistry`-based narrowing also applies to `AuthorizationService.can()` and `AuthorizationService.listObjects()`, so the runtime checks stay type-safe too:
 
 ```ts
-const allowed = await authz.can(user, 'delete', 'post')   // ✓
-const ids = await authz.listObjects(user, 'view', 'post') // ✓
+const allowed = await authz.can(user, 'delete', 'post')     // ✓
+const ids = await authz.listObjects(user, 'delete', 'post') // ✓
 ```
 
 Apps that don't augment `PolicyRegistry` get the loose `(string, string)` fallback — full backwards compatibility.
--- a/docs/guide/authorization.md
+++ b/docs/guide/authorization.md
@@ -41,6 +41,25 @@ new AuthAdapter({
 
 When `roleResolver` is set and `req.tenant` exists, `@Roles()` checks the tenant-scoped roles instead of the user's global roles.
 
+### Type-narrowing roles via `AuthUser` augmentation
+
+`@Roles()` is generic over `AuthUser['roles'][number]`. Augment `AuthUser` to a literal-string array and typos at decoration sites become compile errors — no runtime check needed:
+
+```ts
+declare module '@forinda/kickjs-auth' {
+  interface AuthUser {
+    id: string
+    email: string
+    roles: ('admin' | 'editor' | 'viewer')[]
+  }
+}
+
+@Roles('admin', 'editor')   // ✓ typechecks
+@Roles('typo')              // ✗ TS error: 'typo' is not assignable to '"admin" | "editor" | "viewer"'
+```
+
+Apps that don't augment `AuthUser['roles']` get the loose `string[]` fallback — full backwards compatibility.
+
 ## Policy-Based Authorization
 
 For resource-level permissions ("can this user edit THIS post?"), use policies.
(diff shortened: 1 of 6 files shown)


forinda:
Addressed all 3 Copilot review comments in commit `126d76d`:

**1. `PolicyRegistry` not re-exported from package root**
Added `type PolicyRegistry` to the types export in `packages/auth/src/index.ts`. Without this, augmentation via `declare module '@forinda/kickjs-auth' { interface PolicyRegistry { ... } }` would not merge with the symbol the decorator/service signatures reference, and `dist/index.d.mts` wouldn't surface it.

**2. `Role` type silently resolves to `any` for unaugmented apps**
Real bug — `AuthUser` has `[key: string]: any`, so `AuthUser['roles']` was `any`, which extends every conditional branch. `Role` was therefore `any` and `@Roles(...)` was permissive enough to accept non-strings (the documented `string` fallback never fired).

Fixed with the canonical `IsAny<T>` detection (`0 extends 1 & T`) so the fallback genuinely produces `string` for unaugmented apps. Augmented `roles: ('admin' | 'editor')[]` still narrows to the literal union via `NonNullable<...> + infer T extends string`. Did not touch the `AuthUser` interface so existing apps that augment with non-conventional shapes are unaffected.

**3. `docs/guide/authorization.md` references internal type + uses an action not in the example registry**
- Dropped the `<R extends PolicyResource>` advice (PolicyResource is an internal helper, not exported — consumers can't actually name it).
- Fixed `listObjects(user, 'view', 'post')` → `listObjects(user, 'delete', 'post')` since the example registry above declares `'create' | 'update' | 'delete' | 'publish'` for `'post'` (no `'view'`).

**Verification:**
- `pnpm test` — 159/159 auth tests still pass; the existing `@ts-expect-error` guards in `typed-decorators.test.ts` would have failed the build if the IsAny fix broke either narrowing direction.
- `pnpm typecheck` — green across all 23 packages
- `pnpm format:check` — clean
- `auth` package builds clean (`dist/index.d.mts` regenerated, ~241kB)

PR #184, opened 2026-05-05
feat(db): SQLite relational compiler + Copilot review polish (M4.A.2)

forinda:
## Summary

Closes M4.A.2 from `docs/db/m4-plan.md` — SQLite relational compiler. After this lands, `db.query.X.findMany({ with })` works on PG **and** SQLite. MySQL ships in M4.A.3.

Also bundles 6 Copilot fixes from PR #183 (M4.B, already merged) — small spec/error/runtime corrections that compound naturally on this branch since the M4.A.2 refactor touches `extract-relations.ts` and the relations-snapshot shape.

### M4.A.2 — SQLite compiler

```ts
const db = createDbClient({ schema, dialect: sqliteDialect({ database }) })

const rows = await db.query.users.findMany({
  with: { posts: { with: { comments: true } } },
  where: (_u, eb) => eb('isActive', '=', true),
  limit: 20,
})
//  ^^^ User[] where each user has `posts: (Post & { comments: Comment[] })[]`
```

Compiled SQL uses SQLite's primitives (`coalesce(json_group_array(json_object(...)), '[]')` for `many`, `json_object(...)` with `LIMIT 1` for `one`) — same row-shape contract as PG. `coalesce` keeps the empty-set behavior `[]` not `null`.

**`createDbClient` auto-attaches `ParseJSONResultsPlugin` for SQLite.** SQLite drivers return JSON columns as TEXT; without the plugin, nested `with` results would land as JSON-encoded strings. PG clients skip the plugin (PG decodes natively).

**Refactor: shared traversal.** Extracted from `compile-pg.ts` into `compile-shared.ts`. Both compilers (and MySQL when it lands) are now ~30-line wrappers that pass their helper bag into `runCompile()`. Single source of truth for alias generation, with-walking, where/orderBy/limit/offset plumbing.

**Behavior change:** `buildInnerSelect` emits explicit `.select([col1, col2, ...])` from the snapshot's column list instead of `.selectAll()`. Required for SQLite (Kysely's helpers can't introspect `selectAll()` to build the JSON object's key list); PG's helpers accept either form.

**`CompileFn` signature** gained a `tables` parameter. `createDbClient` users see no change — the snapshot's tables are threaded through `InternalContext.query.tables` automatically. Direct callers of `compilePg` / `compileSqlite` add the new arg.

### Copilot review polish (PR #183 fallout)

| # | Issue | Fix |
| - | ----- | --- |
| 1 | `RelationalQueryAmbiguousRelationNameError` blamed source for duplicates that live on target | Constructor now takes `targetTable` explicitly; message lists conflicting keys with `target.key` prefix |
| 2 | `extract-relations` dropped `rel.relationName` on step 2 / step 3 fallback | Always preserves the source-side tag on resolved `many` entries |
| 3 | Truthy `rel.relationName ?` checks dropped empty-string tags silently | Switched to `!== undefined` |
| 4 | JSDoc on the ambiguous error claimed it fires for `many` duplicates | Corrected — fires only on inverse-`one` ambiguity, scope clarified to `(source, target, name)` |
| 5 | DSL helpers had the same truthy issue | Same `!== undefined` fix |
| 6 | Spec §1 implied two failure modes happened on the same run | Split into Case A (M3 first-match bug) + Case B (FK fallback ambiguity) — mutually exclusive |

Plus spec polish: §2 goal #4 reworded for `with`-key checking accuracy, §4 step 4 error-message claims trimmed to match actual implementation, §5 mapped type uses `relationName?` to match optional contract.

## Suite

| Package | Files | Tests | Δ |
|---|---|---|---|
| `@forinda/kickjs-db` | 54 | **326** | +14 (13 SQLite snapshots + 2 builder tests, −1 SQLite-throws test removed) |
| `@forinda/kickjs-db-pg` | 4 | 23 | unchanged |
| `@forinda/kickjs-cli` | typecheck | unchanged | — |

Build clean. Typecheck clean.

## Test plan

- [x] `pnpm test` green across affected packages.
- [x] `pnpm typecheck` clean.
- [x] `pnpm format:check` clean.
- [x] All 16 PG snapshot fixtures still pass (no SQL output regression after the shared-traversal refactor).
- [x] 13 new SQLite fixtures cover bare findMany, 1-deep many/one, 2-deep, self-reference, per-relation where/limit, outer where+orderBy+limit+offset, mode flags, error paths.
- [ ] Real-SQLite integration test in `packages/db-sqlite/__tests__/integration/` — deferred to M4.A.5 when the peer package lands.


forinda:
## Copilot review addressed in [`429fb7f`](https://github.com/forinda/kick-js/pull/184/commits/429fb7f6)

| # | Comment                                                            | Disposition                                                                                                                                                                                            |
| - | ------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1 | `ParseJSONResultsPlugin` auto-attached for MySQL too                | Gate narrowed to SQLite only. MySQL flips on in M4.A.3 when the real compiler ships — pre-emptive plugin install was misleading.                                                                       |
| 2 | SQLite tests stub `executeQuery`, never exercise the plugin chain  | File header now explicitly notes the limitation. End-to-end JSON-string-to-object round-trip lands in `db-sqlite`'s integration suite (M4.A.5) where a real `better-sqlite3` driver returns the TEXT.   |
| 3 | `raw?: boolean` flag is a no-op                                    | Removed from both `CompileOptions` and `FindManyOptions`. The flag was carried from M3 spec §7 R-1 but never wired. Public API claimed an opt-out adopters couldn't actually use.                       |
| 4 | Changeset documents a public API break for an unexported API        | Reworded — the dialect compilers aren't on the package barrel; signature change is internal. Adopter migration: none.                                                                                  |
| 5 | No explicit empty-string `relationName` test                       | Added in `extract-relations.test.ts`. Locks the `!== undefined` presence-check against future regression to truthy `?` spreads.                                                                        |
| 6 | Stale file header on `query-builder.test.ts`                       | Updated — no longer claims SQLite throws. Now describes actual coverage (SQLite emits dialect-specific SQL; MySQL still throws).                                                                       |

**Suite:** `@forinda/kickjs-db` 54f / **327 tests** (+1 empty-string test). Typecheck clean. Hook gates clean.

Thanks @copilot — caught a misleading pre-emptive plugin install and a no-op flag that would have shipped to v5.4.

PR #190, opened 2026-05-06
fix(kickjs): harden defineContextDecorator (six review-pass tightening)

forinda:
## Why

Code-review pass on `defineContextDecorator`. Six findings — all backwards-compatible tightening, no API surface changes:

| # | Issue | Fix |
| - | ----- | --- |
| 1 | Zero runtime validation of `spec`. Missing `resolve` / empty `key` blew up at request time. | `TypeError` at definition time (typically module load) for missing `resolve`, empty `key`, non-function `onError`, non-array `dependsOn`, non-object `spec`. |
| 2 | No source-location capture on registration. `MissingContributorError` couldn't point adopters at the file that declared the broken decorator. | Capture `new Error().stack` at decorator construction; thread through to `MissingContributorError` message as `declared at src/.../file.ts:L:C`. |
| 3 | Trailing `as unknown as ContextDecorator<...>` double cast. | Replaced with overloaded function signatures + `Object.assign` + `Object.freeze`. Factory return path has no `as unknown` escape hatch anymore. |
| 4 | No `.name` / `.toString` on the returned decorator. | `console.log(LoadTenant)` now prints `[Function: ContextDecorator(tenant)]`. |
| 5 | Stale "No runtime behaviour is wired in Phase 1" JSDoc — Phase 1 shipped. | Replaced with concrete pointer to the new boot-time validation. |
| 6 | Undocumented `({} as D)` cast in `Object.freeze({ ...(spec.deps ?? ({} as D)) })`. | Inline comment explains when the cast is sound, when it isn't, and why the runner's loud-fail behaviour is the right tradeoff. |

The highest-impact fix is #1 — boot-time validation. ~15 lines, catches a whole class of definition-time bugs that previously rode to first request as cryptic ContextMeta misses.

## What changed

**`packages/kickjs/src/core/context-decorator.ts`** — main file. Validation block + stack capture at the top of `defineContextDecorator`. Function declaration changed to overloaded signatures so the return type is structural, not cast. `Object.assign` adds `.registration` + `.with` properties typed via the assign intersection. `Object.freeze` locks the result. `Object.defineProperty` on `.name` for readable stack traces.

**`packages/kickjs/src/core/context-errors.ts`** — `MissingContributorError` gained a fourth optional constructor argument (`dependentDefinedAt?: string`) + a `formatDeclaredAt(stack)` helper that slices the first non-framework frame and formats it as `\n    declared at <frame>` for the message body. Existing call sites continue to work — the parameter is optional and the message falls back to the prior format when absent.

**`packages/kickjs/src/core/contributor-pipeline.ts`** — single-line change: `throw new MissingContributorError(dep, reg.key, route, reg.definedAt)` (added the fourth arg).

**`ContributorRegistration` interface** — gained optional `definedAt?: string` field.

## Tests

- 6 new validation cases on `defineContextDecorator`: null spec, missing/empty `key`, missing/non-function `resolve`, non-function `onError`, non-array `dependsOn`, key-naming in error messages.
- 1 new assertion on `MissingContributorError` that verifies the message includes `declared at` + the test file's path.
- 2 new sanity assertions: `.name` is set to `ContextDecorator(<key>)`; `definedAt` is captured on the registration.

Suite: 366 → 373 tests (+7). Build + typecheck clean.

## Test plan

- [x] `pnpm test` (kickjs) — 373/373 pass
- [x] `pnpm typecheck` (kickjs) — clean
- [x] `pnpm build` (kickjs) — clean
- [x] `pnpm format` — clean

## Summary by CodeRabbit

* **Improvements**
  * Boot-time error messages now include a "declared at" location for missing dependencies, making it easier to see where a required item was registered.
  * Enhanced boot-time diagnostics and stack-capture for clearer, more actionable error reporting.

* **Tests**
  * Expanded tests covering decorator metadata, descriptive naming, and error location tracking.

### DIFF
--- a/.changeset/define-context-decorator-hardening.md
+++ b/.changeset/define-context-decorator-hardening.md
@@ -1,0 +1,21 @@
+---
+'@forinda/kickjs': patch
+---
+
+Harden `defineContextDecorator` based on review feedback. Six tightening passes, all backwards-compatible:
+
+1. **Boot-time spec validation.** `defineContextDecorator` now throws `TypeError` immediately if `spec` is missing/non-object, `spec.key` is empty, `spec.resolve` isn't a function, `spec.onError` is provided but not a function, or `spec.dependsOn` is provided but not an array. Adopters get definition-time errors (typically module load) instead of cryptic ContextMeta misses at first request.
+
+2. **Source-location capture.** Every registration now carries `definedAt: string` — a snapshot of `new Error().stack` taken at decorator-construction time. The contributor pipeline threads it into `MissingContributorError`'s message so boot-time errors print `declared at src/contributors/load-project.ts:42:18` instead of forcing adopters to grep for the key string.
+
+3. **Cleaner type story.** Replaced the trailing `as unknown as ContextDecorator<...>` double-cast with overloaded function signatures + `Object.assign` + `Object.freeze`. `decoratorOrFactory` now matches `ContextDecorator`'s call shapes structurally and properties are typed via the assign intersection — no more `as unknown` escape hatch in the factory's return path.
+
+4. **Meaningful `.name` on the returned decorator.** `console.log(LoadTenant)` now prints `[Function: ContextDecorator(tenant)]` instead of `[Function: decoratorOrFactory]`. Stack traces and devtools inspections name the contributor by its key.
+
+5. **Stale-comment sweep.** Dropped the "No runtime behaviour is wired in Phase 1" line — Phase 1 shipped, the topo-sort + runner + HTTP integration are all live. Replaced with a concrete pointer to the new boot-time validation.
+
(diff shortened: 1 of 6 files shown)


forinda:
Thanks @coderabbitai — addressed in [`5401c81`](https://github.com/forinda/kick-js/pull/190/commits/5401c817).

Added the element-level guard you suggested:

```ts
if (
  Array.isArray(spec.dependsOn) &&
  spec.dependsOn.some((dep) => typeof dep !== 'string' || dep.length === 0)
) {
  throw new TypeError(
    `defineContextDecorator(${spec.key}): spec.dependsOn entries must be non-empty strings.`,
  )
}
```

Plus three test cases:
- `dependsOn: [42]` → throws
- `dependsOn: [''] as never[]` → throws (with the cast since augmented `ContextMetaKey` narrows the entry type once a project augments `ContextMeta`)
- `dependsOn: ['valid', undefined, 'also-valid']` → throws

Catches the slip-through paths (JS callers, `as any` erasures, hand-edited string arrays). Suite 373 → 374 (+1), build + typecheck clean.

PR #191, opened 2026-05-06
feat(kickjs,cli,testing): defineModule factory + simplified routes shape

forinda:
## Why

Three coupled adopter-facing wins:

1. **`defineModule()`** lands as the fourth `define*` primitive — adopters learn one mental model across `defineAdapter` / `definePlugin` / `defineContextDecorator` / `defineModule`. Configurable, scoped, validated at boot.
2. **`ModuleRoutes` simplified**: `routes()` returning `{ path, controller }` is now sufficient — the framework derives the Express Router via `buildRoutes(controller)` internally. `router: buildRoutes(...)` boilerplate is gone (still accepted for hand-built routers).
3. **CLI + docs sweep** pulls every generated module + tutorial onto the new shape so newcomers don't see the old boilerplate.

The legacy `class FooModule implements AppModule { ... }` form keeps working — `bootstrap` accepts either shape and the loader discriminates at boot. No breaking changes.

### `@forinda/kickjs` (minor)

- **New:** `defineModule({ name, defaults, build })` in `core/define-module.ts`. Surface: `(config?)` / `.scoped(scopeName, config?)` / `.definition`. Boot-time validation throws `TypeError` for missing `name` / non-function `build` / non-object options. `.async()` intentionally **not** on the surface — modules have no async-resolution window between `register()` and `routes()`; adopters route async config through an adapter's `beforeStart` instead.
- **New:** `AppModuleEntry = AppModuleClass | AppModule` union — accepted by `bootstrap({ modules })`, `KickPlugin.modules?()`, `createTestApp({ modules })`. The Application loader discriminates `typeof entry === 'function'` to dispatch.
- **Loosened:** `ModuleRoutes.router` is now optional. When omitted, the framework calls `buildRoutes(controller)`. `controller` was already required for OpenAPI introspection, so the simplification removes redundant boilerplate without losing capability.

### `@forinda/kickjs-cli` (minor)

- Every module template (`generateModuleIndex` DDD, `generateRestModuleIndex`, `generateMinimalModuleIndex`, `cqrs.ts`'s `generateCqrsModuleIndex`, `scaffold.ts`'s `genModuleIndex`, `project-app.ts`'s `generateHelloModule`) emits the `defineModule({ name, build })` form with `{ path, controller }` route shape.
- Generated `routes()` carries a JSDoc demonstrating array-return + per-entry `version` override so adopters discover that surface from the file:

  ```ts
  /**
   * Return an array to mount multiple route sets — each entry can
   * override the API version with a `version` field:
[...]

PR #193, opened 2026-05-07
feat(cli): kick.config.ts > modules.style ('define' | 'class')

forinda:
## Why

Project hygiene flag for codegen. After [#191](https://github.com/forinda/kick-js/pull/191), `kick g module` and `kick g scaffold` always emit the new `defineModule({...})` factory shape. That's the right default for new code, but adopters with established class-style codebases want the orchestrator to keep emitting matching shapes — mixed styles in one repo are visible drift. This PR adds the opt-out lever.

## What's new

```ts
// kick.config.ts
export default defineConfig({
  modules: {
    style: 'class', // pin to legacy form; default is 'define'
  },
})
```

| Style | Module file | Modules registry |
| --- | --- | --- |
| `'define'` (default) | `defineModule({ name, build: () => ({...}) })` | `[TaskModule()]` |
| `'class'` | `class TaskModule implements AppModule { ... }` | `[TaskModule]` |

The framework runtime accepts both shapes regardless of this setting — `Application` discriminates `typeof entry === 'function'` at boot. The flag controls codegen output only.

`kick rm module` matches both `Module` and `Module()` forms in the modules array, so flipping the flag mid-project doesn't break un-registration. Mixing styles in the same project works.

## What changed

- **`ModuleConfig.style?: 'define' | 'class'`** added to the `kick.config.ts` schema. Unknown values warn + fall back to `'define'`.
- All four pattern generators (DDD / REST / CQRS / minimal) + `kick g scaffold`'s template branch on the resolved style.
- `kick g module` orchestrator emits `Module()` (factory call) for `'define'`, bare `Module` for `'class'`.
- `kick g scaffold`'s auto-register threads style through the same way.
- `ModuleStyle` type exported from `templates/types.ts` and threaded via `ModuleContext`.

## Tests

- `kick g module` (no config) → emits `defineModule(...)`, registry entry is `TaskModule()`.
- `kick g module` with `kick.config.json: { modules: { style: 'class' } }` → emits `class TaskModule implements AppModule`, registry entry is bare `TaskModule`.
- One existing scaffold test updated for the simplified `ModuleRoutes` shape (#191 dropped the redundant `router: buildRoutes(X)` field).

Suite: 231 → 234 (+3). Build + typecheck clean.

## Docs

`docs/guide/generators.md` "Module declaration style" section added under "Config-Aware Defaults" — covers the flag's effect on both the module file and the registry, plus the runtime guarantees.

## Related — migration command (deferred)

A future `kick migrate modules --experimental` command will rewrite class-form modules to the `defineModule` factory form via TypeScript AST transforms. Out of scope here; the config flag covers the inverse direction (pin to class form) for projects that don't want to migrate.

## Test plan

- [x] `pnpm test` (CLI) — 234/234 pass
- [x] `pnpm typecheck` (CLI) — clean
- [x] `pnpm build` (CLI) — clean
- [x] `pnpm format` — clean

## Summary by CodeRabbit

* **New Features**
  * Configurable modules.style ('define' | 'class') that controls generated module shape, scaffold output, and generator gating.
  * New experimental codemod (kick codemod modules) to migrate module declarations with dry-run/--apply, timestamped backups, idempotent behavior, and conservative skipping.

* **Documentation**
  * Added Module declaration style section to generator docs and CLI docs for migration/gating guidance.

* **Tests**
  * Expanded E2E/unit tests covering generation, scaffolding, migration, gate behavior, idempotency, backups, and discovery rules.

### DIFF
--- a/.changeset/cli-module-style-config.md
+++ b/.changeset/cli-module-style-config.md
@@ -1,0 +1,34 @@
+---
+'@forinda/kickjs-cli': minor
+---
+
+`kick.config.ts > modules.style: 'define' | 'class'` — opt-out flag for projects that prefer the legacy `class FooModule implements AppModule { ... }` form over the new `defineModule({...})` factory.
+
+```ts
+// kick.config.ts
+export default defineConfig({
+  modules: {
+    style: 'class', // pin to legacy form; default is 'define'
+  },
+})
+```
+
+The framework runtime accepts both shapes regardless of this setting — the flag controls codegen output only:
+
+- **`'define'`** (default) — `defineModule({ name, build: () => ({...}) })` factory form, with the orchestrator inserting `TaskModule()` (factory call) into `src/modules/index.ts`.
+- **`'class'`** — legacy `class FooModule implements AppModule { ... }`, with the orchestrator inserting bare `TaskModule` (no call) into the modules array.
+
+`kick rm module` matches both `Module` and `Module()` forms, so a project that flips the flag mid-stream can keep adding/removing modules without breakage. Mixing styles in the same project works (loader discriminates `typeof entry === 'function'` at boot).
+
+## What changed
+
+- `ModuleConfig.style?: 'define' | 'class'` added to `kick.config.ts` schema. Unknown values warn + fall back to `'define'`.
+- All four pattern generators (DDD, REST, CQRS, minimal) + the `kick g scaffold` template branch on the resolved style.
+- `kick g module`'s array-insertion regex emits `Module()` for `'define'` and bare `Module` for `'class'`.
+- New tests cover both branches: `kick g module` with no config (default 'define') AND with `kick.config.json: { modules: { style: 'class' } }`.
+
+Suite: 231 → 234 (+3 new style tests). Build + typecheck clean.
+
+## Related — migration command (deferred)
+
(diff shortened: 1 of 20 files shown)


forinda:
Thanks @coderabbitai — all four addressed in [`7a45f54`](https://github.com/forinda/kick-js/pull/193/commits/7a45f546).

| # | Fix | Commit hunk |
| - | --- | --- |
| 1 | Fence missing language tag in changeset's example output | Added `text` after the opening fence |
| 2 | `isDryRun` ignored top-level `--dry-run` | Switched to `cmd.optsWithGlobals()` so the global flag chain resolves correctly. `kick --dry-run codemod modules --apply` now correctly resolves to dry-run |
| 3 | `migrateDefineToClass` emitted `contributors(): ContributorRegistrations` without importing the type | Extended `rewriteImportsForClass` with a `contributorRegistrations` flag in the `needs` object; set when the source defineModule had a `contributors()` block. Migrated files no longer fail tsc on this path. |
| 4 | `findModuleFiles` swept barrel files at every depth | Replaced the `isRoot` boolean with a depth counter; legacy `index.ts` discovery now constrained to `depth === 1` (`<modulesDir>/<sub>/index.ts`), matching the historical convention exactly. Nested barrel files like `<sub>/application/index.ts` are correctly excluded. |

Bonus fifth (your last comment on the registry-only backup): backup snapshot now also fires when `moduleFiles.length === 0` but the registry exists, so registry-only rewrites get the same safety net the command advertises.

Three new tests cover #3 (contributors() import path), #4 (depth-aware discovery against a nested-barrel fixture), and the registry-only backup case. Suite 253 → 256.

PR #484, opened 2026-07-25
feat: add @forinda/kickjs-grpc — Connect RPC transport

forinda:
## Why

KickJS had no RPC story. The [roadmap](https://kickjs.app/guide/roadmap) rejects a microservices module but carves out the path explicitly:

> If we want a story here, it's a dedicated adapter package, not a core feature.

This is that package. It follows the `WsAdapter` sidecar precedent rather than pretending gRPC is a fourth HTTP runtime.

## What it does

Serves **gRPC-Web, the Connect protocol, and (behind an HTTP/2 terminator) native gRPC from the same app and the same port** as your HTTP routes.

```ts
@GrpcService(UserService)          // generated protobuf DescService
export class UserRpc {
  @Autowired() private users!: UserRepository

  @GrpcMethod()
  async getUser(req: GetUserRequest, ctx: GrpcContext) {
    const user = await this.users.findById(req.id)
    if (!user) throw HttpException.notFound(`No user ${req.id}`)   // → Code.NotFound
    return { id: user.id, email: user.email }
  }
}

bootstrap({ modules: [UserModule()], adapters: [GrpcAdapter()] })
```

Smoke-testable with no client codegen, because the Connect protocol is JSON over HTTP:

```bash
curl -X POST http://localhost:3000/user.v1.UserService/GetUser \
  -H 'Content-Type: application/json' -d '{"id":"1"}'
```

## Design decisions worth reviewing

**Not an `HttpRuntime`.** That seam is committed to `nodeHandler(app): (req: IncomingMessage, res: ServerResponse) => void` and `RouteEntry { method: 'GET'|…, path: '/x/:id' }`. gRPC is `/package.Service/Method` over HTTP/2 framing with length-prefixed protobuf — modelling it as a runtime would mean lying about the contract. It is a peer transport alongside ws/queue.

**Mounted at the `beforeGlobal` middleware phase.** This is the load-bearing choice. That phase runs *after* in-flight request tracking and the request-scope ALS frame, but *before* the global middleware stack — so Connect gets the raw request body it must parse itself. Mounting any later means `express.json()` drains the stream first and every RPC hangs. Mounting at `beforeMount` instead would work but would skip in-flight tracking. Connect's `fallback` hands non-RPC requests back to `next()`, so HTTP routes on the same port are untouched.

**Contributors came almost free.** `runContributors()` already took `ExecutionContext`, not `RequestContext`, and `execution-context.ts` says so outright: *"`WsContext`, `QueueContext`, and `CronContext` will adopt it in V2."* `GrpcContext implements ExecutionContext`, so `defineContextDecorator` works on RPCs unchanged — same `dependsOn` topo-sort, same **method > class > adapter** precedence, same boot-time cycle detection.

**Streaming returns an async generator synchronously.** An `async` wrapper would hand Connect a `Promise<AsyncIterable>` instead of an `AsyncIterable`. The generator form still lets the contributor pipeline `await` before the first message is yielded.

**Boot-time failures over request-time surprises.** Unknown RPC names and contributor cycles abort startup. RPCs declared in the proto but not implemented are *not* an error — Connect answers `Code.Unimplemented`.

## Honest limitation, documented prominently

The adapter rides the KickJS HTTP server, which speaks HTTP/1.1:

| Protocol | Works | Notes |
| --- | --- | --- |
| Connect | ✅ | JSON or binary over HTTP/1.1. Curl-able. |
| gRPC-Web | ✅ | Browser clients. No proxy needed. |
| native gRPC | ❌ | Needs HTTP/2 **and** HTTP trailers. |

Native gRPC requires an HTTP/2 terminator in front (Envoy, nginx, a mesh, Cloud Run). This is stated in the README, the guide, the API page, and the adapter's TSDoc — not glossed.

This is also why `prefix` defaults to empty: most native gRPC clients hard-code the path and cannot send a prefix. Consequence worth noting in review — **RPCs sit at the root**, not under `apiPrefix`, while HTTP routes keep `/api/v1/...`.

## Error mapping

`HttpException` maps by status; a `ConnectError` passes through untouched; anything else becomes `Code.Internal` with the original preserved as `cause` (logged, never sent on the wire). `onError` can rewrite anything.

| HTTP | Connect code | HTTP | Connect code |
| --- | --- | --- | --- |
| 400, 422 | `InvalidArgument` | 429, 413 | `ResourceExhausted` |
| 401 | `Unauthenticated` | 500 | `Internal` |
| 403 | `PermissionDenied` | 501, 405 | `Unimplemented` |
| 404 | `NotFound` | 503 | `Unavailable` |
| 409 | `AlreadyExists` | 504, 408 | `DeadlineExceeded` |

## Testing

**21 tests**, including a real `http.Server` integration suite that proves the shared-port claim rather than asserting it: unary RPC, server-streaming, error-code propagation, a raw JSON `fetch`, an HTTP route on the same port, and 404 fall-through.

The protobuf fixture is built at runtime via `createFileRegistry`, so the suite needs no `buf generate` step and no checked-in codegen output.

Verified green:

- `pnpm build` — 26/26 tasks
- `pnpm lint` (oxlint), `pnpm lint:tokens`, `pnpm format:check`
- `pnpm docs:build` — clean, no dead links
- `@forinda/kickjs` (703), `-ws`, `-swagger`, `-grpc` (21) all pass

## Two pre-existing issues found while verifying

Neither is caused by this PR; both reproduce on a clean stashed baseline.

1. **`packages/vscode-extension` build fails** — stale `vite@8.0.16_@types+node@25.*` link, `MODULE_NOT_FOUND`. A full `pnpm install` did not clear it. Worth a separate look.
2. **Full-concurrency `turbo run test` intermittently fails `@forinda/kickjs#test` and `@forinda/kickjs-db-pg#test`** on a local Windows machine. Both pass in isolation (703/703, 74/74) — the db suites are testcontainers-backed and heavy. Resource contention, not a regression.

## Dependencies

`@connectrpc/connect`, `@connectrpc/connect-node`, and `@bufbuild/protobuf` are **peer** dependencies, so adopters control the protobuf runtime version. The lockfile diff is purely additive — only this package's entries.

---

https://claude.ai/code/session_01TATbsR29p9nqAV9KBRKm1L

## Summary by CodeRabbit

* **New Features**
  * Added gRPC, gRPC-Web, and Connect RPC support through a new KickJS package.
  * Supports unary and streaming RPCs alongside existing HTTP routes on the same server and port.
  * Added decorator-based service and method registration, RPC context access, error mapping, and runtime service statistics.
  * Added support for request metadata, deadlines, cancellation, contributors, and configurable protocol and routing options.

* **Documentation**
  * Added comprehensive API, setup, configuration, and protocol documentation for the new package.


forinda:
Thanks — all three addressed in 166f94fd. One of them I implemented differently from the suggestion; reasoning below.

---

### 🟠 Major — "Do not serialize unexpected internal error messages"

**Fixed, and you were right that this was a real defect rather than a preference.** The docstring on `toConnectError` already claimed the behaviour you asked for — *"preserving the original as `cause` for logging without leaking it onto the wire"* — while the code did the opposite. Code and contract disagreed; the contract was the correct one.

Unexpected throws (plain `Error`, `MissingContextValueError`, non-`Error` values) now respond with an exported `INTERNAL_ERROR_MESSAGE` constant, with the original kept on `cause`.

Two things I added beyond the literal suggestion:

1. **Server-side logging.** Redacting alone would have made the failure vanish entirely — `cause` isn't surfaced by connect-node, so a dropped DB connection would have produced a bare `Internal error` and nothing in the logs. `fail()` now logs `Code.Internal` responses at `error` level with the RPC name. Deliberate faults (`ConnectError` / `HttpException`) already describe themselves to the caller and don't get an error-level log.

2. **A latent bug in the `HttpException` branch.** It passed `err.details` as `ConnectError`'s 5th parameter (`cause`). Connect's 4th parameter is `OutgoingDetail[]` — protobuf messages — which arbitrary JSON validation details cannot satisfy, so those details were only ever reachable through `cause` anyway. It now passes the exception itself, so `cause` carries the message *and* the details.

`HttpException` and `ConnectError` messages still reach the client unchanged — raising either is a deliberate act of describing a fault to the caller, exactly as it would be over HTTP. Redacting those would break the documented mapping table.

Four new tests, including an end-to-end one asserting a `connect ECONNREFUSED 10.0.0.5:5432` message does not cross the wire.

---

### 🟡 Minor — "Use the complete contributor precedence contract everywhere"

**Partially disagreeing, and I'd like a second opinion if you think I've got this wrong.**

Restating `method > class > module > adapter > global` for this transport would have been *inaccurate*. This adapter implements **three of those five**, and the two it omits are omitted by construction, not by oversight:

- `AppModule.contributors()` is collected inside the module **route**-mounting loop in `application.ts` and passed to `buildRouteTable` as `externalSources`.
- `bootstrap({ contributors })` becomes `globalSources` in the same place.

Both feed the HTTP route table. Neither is reachable from a non-HTTP transport — I verified this in `application.ts` rather than assuming it. Documenting the full five-level chain would have promised adopters that `bootstrap({ contributors })` covers their RPCs, which would fail silently at runtime — a worse outcome than the incomplete wording.

So I've documented **both**: the framework's full chain *and* which levels this transport participates in, with the reason. Applied in all four locations you flagged, plus the `GrpcAdapterOptions.contributors` docstring:

> The framework's full precedence chain is **method > class > module > adapter > global**. RPCs participate in three of those: **method > class > adapter**. The `module` and `global` levels are HTTP-only by construction […] To cover every RPC, register through `GrpcAdapter({ contributors })` instead.

If the guideline is meant to mandate the literal string regardless of what a given transport implements, say so and I'll reconsider — but I'd rather the docs be true than uniform.

---

### 🟡 Minor — "Keep quick-start snippets self-contained"

**Fixed.** Added `UserRepository` + `GetUserRequest` to `docs/guide/grpc.md`, and `HttpException` + `GetUserRequest` to `packages/grpc/README.md`.

---

**Verification:** 25 tests pass (21 → 25), `tsgo --noEmit` clean, `pnpm build`, `pnpm docs:build`, `oxfmt --check`, `oxlint`, and `kick-lint` all green.

PR #487, opened 2026-07-25
chore: adopt TypeScript 7; fix baseUrl removal in test tsconfigs

forinda:
## Why

[TypeScript 7.0](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/) shipped 2026-07-08. `tsc` **is** the native Go binary now, so the `@typescript/native-preview` side-install and the `tsgo`/`tsc` split this repo carried are no longer needed.

The headline blocker in that announcement — **TS 7.0 ships no compiler API**, stranding `ts-loader`, Vue/Astro/Svelte tooling, and `typescript-eslint` until 7.1 — does not apply here. Nothing in the repo imports `typescript`: typegen parses with `oxc-parser`, linting is `oxlint`, bundling is `rolldown`. I swept for `from 'typescript'` / `require('typescript')` repo-wide and got zero hits.

## Two pre-existing bugs this surfaced

**1. 19 `tsconfig.test.json` files were dead config.** They set `baseUrl`, which TS 7 removed — and the *already-pinned* native-preview rejected them:

```
error TS5102: Option 'baseUrl' has been removed.
error TS5090: Non-relative paths are not allowed. Did you forget a leading './'?
```

They went unnoticed because `pnpm typecheck` reads `tsconfig.json` and vitest typecheck is off by default. Removing `baseUrl` alone wasn't enough — the `paths` *values* also had to become relative.

`packages/db` is the one package that genuinely runs `vitest --typecheck`, and its config had neither problem, so **nothing that executes today changes behaviour**.

**2. The CLI test helper resolved the `tsgo` binary by name.** Dropping native-preview broke 26 generator tests with `tsgo binary not found`. It now prefers `tsc`, keeping `tsgo` in the probe list for trees that still have the preview package.

## ⚠️ Scope call: I did not fix the newly-visible test-type errors

Making those 19 configs *load* reveals pre-existing type errors in test code across ~13 packages — implicit `any`, possibly-undefined invocations, plus two genuine config gaps (`packages/lint` needs `types: ["node"]`; `packages/devtools-kit` needs the `dom` lib).

**None are fixed here.** They were invisible before this PR and remain unexecuted after it. Fixing them is a real chunk of work that belongs in its own PR rather than smuggled into a toolchain change — but it's now a tracked, visible follow-up instead of a silent one.

## The upgrade

| Scope | Change |
| --- | --- |
| root | `typescript` `^6.0.3` → `^7.0.2`; `tsdown` `^0.22.3` → `^0.22.14`; `@typescript/native-preview` **removed** |
| 23 packages | `typecheck` script `tsgo --noEmit` → `tsc --noEmit`; `typescript` devDep → `^7.0.2` |
| 23 tsdown configs | `dts: { tsgo: true }` → `dts: true` |
| CLI | `kick new` scaffolds `typescript: ^7.0.2` |

**`tsdown` had to move too.** 0.22.3 pulled `rolldown-plugin-dts@0.26`, whose peer range caps TypeScript at `^6`. 0.22.14 pulls 0.27, which accepts `~7.0.0`. That was the actual blocker — not the API.

**`dts: true` is not a downgrade.** `rolldown-plugin-dts` auto-selects the `tsgo` generator when TypeScript 7 is installed; the explicit flag is now redundant and would otherwise point at the removed preview package.

`tsconfig.base.json` needed **no changes** — `target: ES2023`, `module: ESNext`, `moduleResolution: bundler`, explicit `types`, and `strict: true` were already TS 7-clean, and none of the removed options (`target: es5`, `downlevelIteration`, `node10`/`classic` resolution, AMD/UMD/System) appear anywhere.

## Decision worth reviewing: scaffolded projects

`kick new` now pins `typescript: ^7.0.2` (changeset: `@forinda/kickjs-cli` **minor**). The generated `tsconfig.json` already used `moduleResolution: 'bundler'` with no `baseUrl`, so it needed no changes — and the generator suite typechecks every scaffolded fixture with `tsc --noEmit`, passing unmodified against 7.0.2.

The caveat is in the changeset: an adopter who has added `ts-loader`, Vue/Astro/Svelte tooling, or `typescript-eslint` needs to stay on `typescript@6` until 7.1. A stock KickJS project uses none of those. **If you'd rather not push TS 7 onto new projects yet, that's a one-line revert** in `project-config.ts` plus dropping the changeset.

I left the CLI's `kick dev --typecheck` probe (tsgo-then-tsc for the *user's* project) alone — it stays correct either way, since both are native.

## Verification

- `typecheck` **30/30**, `build` **26/26**, `test` **50/50 tasks**, `docs:build` clean
- `oxfmt --check`, `oxlint`, `kick-lint` clean
- **Emitted declarations are byte-identical** — `packages/grpc/dist/index.d.mts` diffs to zero lines against the pre-upgrade build. The aggregate hash over all 105 `.d.mts` files moves only because content-hashed *chunk filenames* changed (`errors-n0H96eEz` → `errors-By6UueCm`); the `.mjs` chunk hashes are unchanged.
- Typecheck was re-run **after** rebuilding, so every package's source is verified against its dependencies' freshly TS7-emitted `.d.mts`, not stale TS6 output.

One note on methodology: my first attempt at a baseline comparison was invalid — `git stash` reverts source but not `node_modules`, so the "TS 6 build" was still running TS 7. The byte-identical result above is from a valid comparison.

---

https://claude.ai/code/session_01TATbsR29p9nqAV9KBRKm1L

## Summary by CodeRabbit

* **New Features**
  * New projects created with `kick new` now use TypeScript 7.0.2.
  * TypeScript builds and checks now use the native compiler, simplifying setup and removing the need for an additional preview compiler package.
* **Documentation**
  * Added upgrade guidance for TypeScript 7, including compatibility considerations for tools that rely on the compiler API.
  * Updated project setup and contribution instructions to reflect the current TypeScript toolchain.


forinda:
Fixed in 0c93eaf0 — you're right, and the wording was worse than the code.

"Cannot run on 7.0 and needs 7.1" framed 7.1 as the destination and TypeScript 6 as a waiting room. Neither is accurate: TS 6 is supported, and the TypeScript team ships `@typescript/typescript6` precisely so a project can run 7 for the build while pointing API-dependent tooling at 6 — which my original text omitted entirely.

I verified the package before citing it rather than taking the announcement's word for it: `@typescript/typescript6@6.0.2`, `bin: { tsc6: 'bin/tsc6' }`.

The guidance now lists all three paths as peers, with 7.1 described as the *expected* future API rather than a required one:

> - Stay on `typescript@6`. It remains supported; this is a steady state, not a stopgap.
> - Install `@typescript/typescript6` alongside 7 and point that tooling at the `tsc6` binary it provides, so the rest of your build gets the native compiler.
> - Wait for the new API, which the TypeScript team expects to ship in 7.1.

Changeset-only change; no source touched.

PR #493, opened 2026-07-27
fix(cli): resolve sibling package versions on Windows instead of falling back to the CLI version

forinda:
## The bug

Projects created by `kick new` pin every `@forinda/kickjs*` dependency to the **CLI's** version rather than each package's own published version.

A scaffold with `--packages swagger,ws,queue,devtools` produced:

```jsonc
"@forinda/kickjs":          "^6.9.1",  // actually published: 6.5.1
"@forinda/kickjs-schema":   "^6.9.1",  // actually published: 0.1.3
"@forinda/kickjs-swagger":  "^6.9.1",  // actually published: 7.1.0
"@forinda/kickjs-ws":       "^6.9.1",  // actually published: 7.0.1
"@forinda/kickjs-queue":    "^6.9.1",  // actually published: 7.0.1
"@forinda/kickjs-devtools": "^6.9.1",  // actually published: 7.1.1
"@forinda/kickjs-cli":      "^6.9.1",  // correct by coincidence
"@forinda/kickjs-vite":     "^6.9.1",  // actually published: 8.0.0
```

`@forinda/kickjs-schema@^6.9.1` doesn't exist at all, so `install` fails outright; the rest silently under- or over-install.

## Root cause

`resolveSiblingVersions()` shells out to `npm view <pkg> version` and falls back to the CLI's own version when the query fails. On Windows the query **always** failed:

```
execFileSync('npm', …)      → ENOENT   // there is no npm.exe — npm is a .cmd shim
execFileSync('npm.cmd', …)  → EINVAL   // Node >=18.20 refuses to spawn batch files
                                       // without a shell (CVE-2024-27980 hardening)
```

The error was swallowed in a bare `catch`, so a command that **never ran** was indistinguishable from a package that **isn't published** — and every dependency took the fallback path at once. The fallback itself is correct and worth keeping (it makes the scaffold work offline); it was just firing 100% of the time on Windows.

Two adjacent helpers had the same defect and were failing silently for the same reason:

- `resolveVersionAtTag()` / `tagExportsSubpath()` — mis-gated the alpha runtime pin, so a `--runtime fastify` / `--runtime h3` scaffold couldn't tell whether the engine subpath had graduated to stable.
- `initFullstackProject()` — the workspace root `<pm> install` used `execFileSync(packageManager, …)`, which fails on Windows for **every** package manager, since `pnpm` / `npm` / `yarn` / `bun` are all `.cmd` shims.

## The fix

Adds cross-platform `captureCommand` / `runCommand` helpers to `packages/cli/src/utils/shell.ts`, routing `.cmd` shims through `cmd.exe /d /s /c` on Windows and calling `execFileSync` directly elsewhere.

`shell: true` would also have worked, but it's rejected here on two counts: it concatenates rather than escapes arguments (Node 24 now emits `DEP0190` warning about exactly this), and that warning would surface as noise in the CLI's own output. Instead the invocation is built explicitly and every argument is checked against a shell-metacharacter guard — package names and dist-tags can never be reinterpreted as `cmd.exe` syntax, and an argument that looks like it might be is refused rather than run.

## After

```jsonc
"@forinda/kickjs":          "^6.5.1",
"@forinda/kickjs-schema":   "^0.1.3",
"@forinda/kickjs-swagger":  "^7.1.0",
"@forinda/kickjs-ws":       "^7.0.1",
"@forinda/kickjs-queue":    "^7.0.1",
"@forinda/kickjs-devtools": "^7.1.1",
"@forinda/kickjs-cli":      "^6.9.1",
"@forinda/kickjs-vite":     "^8.0.0"
```

## Testing

- New `packages/cli/__tests__/shell-capture.test.ts` (5 tests) — exercises the real `npm` binary, which on Windows *is* the `.cmd` shim that used to fail, plus the metacharacter guard, missing binaries, and non-zero exits. Meaningful on POSIX runners too, just not as a platform regression guard.
- Existing scaffold / package-manager / info suites: 31 passing.
- `tsc --noEmit` and `oxlint` clean.
- End-to-end: scaffolded both a default and a `--packages swagger,ws,queue,devtools` project and diffed the generated `package.json` (output above).

## Notes

- Unix behaviour is unchanged — `shellSafeInvocation` returns `[file, args]` untouched off Windows.
- The offline fallback to the CLI version is deliberately retained; it now fires only when `npm` genuinely can't answer.
- Unrelated: `pnpm format` currently rewrites 18 markdown files that `format:check` flags (oxfmt drift predating this branch). Left out of this diff to keep it scoped — worth a separate `style:` commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `kick new` so generated projects resolve each sibling package to its published version instead of always using the CLI version.
  - Fixed Windows failures when installing dependencies or running package-manager commands, including `kick new --template fullstack`.
  - Improved command safety by rejecting unsafe shell arguments and handling failed or unavailable commands consistently.

### DIFF
--- a/.changeset/stale-donuts-repeat.md
+++ b/.changeset/stale-donuts-repeat.md
@@ -1,0 +1,9 @@
+---
+'@forinda/kickjs-cli': patch
+---
+
+fix: scaffolded projects no longer pin every `@forinda/kickjs*` dependency to the CLI's own version
+
+`kick new` resolves each sibling package's published version with `npm view <pkg> version`, falling back to the CLI's version when the query fails. On Windows that query always failed: `npm` is a `.cmd` batch shim, so `execFileSync('npm', …)` raised `ENOENT` (there is no `npm.exe`) and `npm.cmd` raised `EINVAL` (Node >= 18.20 refuses to spawn batch files without a shell — CVE-2024-27980). The error was swallowed, so every generated `package.json` silently collapsed onto the CLI version — `@forinda/kickjs`, `-schema`, `-vite`, `-swagger` and friends all stamped `^<cli version>` even though per-package independent versioning means they diverge.
+
+Version resolution now goes through a cross-platform `captureCommand` helper that routes `.cmd` shims via `cmd.exe` on Windows, so each dependency gets its real published range. The same helper fixes `kick new --template fullstack`, whose root `<pm> install` step failed on Windows for the same reason.
--- a/packages/cli/__tests__/shell-capture.test.ts
+++ b/packages/cli/__tests__/shell-capture.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'vitest'
+import { captureCommand } from '../src/utils/shell'
+
+/**
+ * Regression guard for the "scaffolded projects pin every sibling to the
+ * CLI's own version" bug.
+ *
+ * `resolveSiblingVersions()` shells out to `npm view <pkg> version` and
+ * falls back to the CLI's version whenever the query fails. On Windows
+ * that query ALWAYS failed — `npm` is a `.cmd` batch shim, so
+ * `execFileSync('npm', …)` raised ENOENT (no `npm.exe`) and
+ * `execFileSync('npm.cmd', …)` raised EINVAL (Node >= 18.20 refuses to
+ * spawn batch files without a shell — CVE-2024-27980). Because the
+ * caller swallowed the error, every generated `package.json` silently
+ * collapsed onto the CLI version instead of each package's real one.
+ *
+ * These tests exercise the real `npm` binary on whatever platform they
+ * run on, which is exactly the case that used to break.
+ */
+describe('captureCommand', () => {
+  it('runs a package-manager shim and captures its stdout', () => {
+    // `npm --version` is offline, fast, and — critically — is the `.cmd`
+    // shim on Windows that plain execFileSync cannot launch.
+    const out = captureCommand('npm', ['--version'], { timeout: 30_000 })
+    expect(out).toMatch(/^\d+\.\d+\.\d+/)
+  })
+
+  it('captures multi-token arguments without a shell rewriting them', () => {
+    const out = captureCommand('npm', ['config', 'get', 'registry'], { timeout: 30_000 })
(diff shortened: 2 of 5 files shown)


forinda:
Both findings were valid — fixed in c4535b6e.

**Untestable guard.** Correct, and the diagnosis of *why* was the useful part: on a Linux/macOS runner my metacharacter test passed for the wrong reason. `shellSafeInvocation` returns the invocation untouched off Windows, so `captureCommand('npm', ['view', 'foo && echo pwned', 'version'])` actually ran npm, npm rejected the bogus spec, and the `toBeNull()` assertion held without the guard ever executing. A regression would have sailed through CI green.

`shellSafeInvocation` is now exported and takes the platform as a parameter (defaulted, read per call rather than captured at import time), so both branches are exercised on every host. Added 8 direct tests covering the passthrough branch, the cmd.exe branch, quoting, each rejected metacharacter, a metacharacter in the command name, and a positive case asserting the argument shapes the CLI actually passes still get through.

**Backslashes.** Confirmed against real `cmd.exe` before changing anything, and it's worse than just the odd-trailing-backslash case:

```
['C:\some path\']  ->  child receives  ['"C:\some', 'path\"']
['no-space\']      ->  child receives  ['no-space\']
```

The first splits one argument into two *and* leaks quotes; the second shows Node's argv encoder doubling a trailing backslash even with no space involved. Went with rejecting `\` rather than escaping it: correct escaping is fiddly across the two parsing layers, and no caller needs it — arguments here are package names, dist-tags, subcommands and flags, and working directories travel via the `cwd` option, not as arguments. A loud `null` beats a silently corrupted argument, which is precisely the failure mode this module exists to eliminate.

One correction on severity: this is a correctness bug, not a security one. Reaching command *syntax* still requires a character from the first group (`& | < > ^ " ' \` ( ) [ ] { } ; ! %` or a newline), all of which were already rejected. Worst case pre-fix was a malformed argument, not injection.

Verified the backslash test is non-vacuous by mutation — dropping `\` from the character class fails it (`1 failed | 10 passed`) — and re-ran the end-to-end scaffold after the stricter guard to confirm real version resolution is unaffected.

PR #494, opened 2026-07-28
fix(cli): scaffold return-value controllers so response types actually infer

forinda:
## The bug

`kick typegen` fills `KickRoutes[...].response` with `InferHandlerResponse<Controller['method']>`, which reads the handler's **return type** and nothing else (`packages/kickjs/src/http/reply.ts:64`). Every controller the CLI scaffolded wrote its response imperatively, so adopters got no response typing — and one case was worse than nothing.

Measured by compiling the generated shapes against `InferHandlerResponse`:

| Generated handler          | Inferred `response` (before)                          |
| -------------------------- | ----------------------------------------------------- |
| `ctx.json(result)`         | `unknown`                                              |
| `ctx.created(result)`      | `unknown`                                              |
| `ctx.noContent()`          | `unknown`                                              |
| `return ctx.notFound(...)` | **`RuntimeResponse`** — the framework's internal response driver |

The last row is the sharp one. `ctx.json` / `ctx.notFound` return the runtime's response driver for fluent chaining (`context.ts:672`), so `return ctx.notFound(...)` types the route as that internal object rather than the resource. `routes.ts:152` already documented the softer half of this — *"imperative ctx.json degrades to unknown"* — but the scaffold was hitting the harder half. It was also calling `ctx.notFound()`, deprecated since `ctx.problem.notFound()` landed.

Net effect: the typed client could not know the shape of any route the CLI generated, which is the whole point of the R1/R2 work in `response-inference-design.md`.

## The fix

Controllers now:

- **return** their payload instead of calling `ctx.json()`
- use `reply.created()` / `reply.noContent()` for non-200 statuses — `Reply<S, T>` unwraps to `T`, so the status is carried without losing the payload type
- send error branches through `ctx.problem.*` (RFC 9457) followed by a bare `return`

That last point is the load-bearing detail: the error branch contributes `undefined`, which `InferHandlerResponse` drops, so a 404 no longer widens the success type into a union. Non-2xx already reaches the typed client as a `KickClientError` carrying the problem body, so it belongs on the error channel, not in the response type.

After: `getById` infers the entity, `create` the created entity, `remove` `undefined`.

Covers `kick g module` (rest + minimal), `kick g controller`, `kick g scaffold`, and the `kick g contributor --type http` usage example. The orphaned `auth-scaffold.ts` templates have the same pattern but zero callers repo-wide, so they are left alone.

## Minimal pattern now scaffolds full CRUD

Per request in review: `--pattern minimal` emitted only `list`. It now emits the whole CRUD surface — `list` / `getById` / `create` / `update` / `remove` — in its single controller file. "Minimal" refers to the file count (no service/DTO/repository split), not the route surface. Same for `kick g controller`, which was stuck at `list` + `create`.

## Docs

The controllers guide led with the imperative form in its headline example and buried "Return-Value Handlers" 200 lines down — exactly backwards now that it is what the CLI emits. Inverted, plus:

- an inference table covering all four styles, including the `return ctx.json(x)` → `RuntimeResponse` trap
- an "Error branches" section showing the `ctx.problem.*` + bare `return` pattern
- response-helper table marks `ctx.notFound` / `ctx.badRequest` deprecated in favour of `ctx.problem.*`

`README.md`'s hello controller used `ctx.json(...)` while the actual `generateHelloController` template already emitted return-style — the README contradicted the CLI it documents.

## Testing

New `packages/cli/__tests__/controller-response-inference.test.ts` asserts at the **type level**, compiled by the fixture's own `tsc`. This is deliberate: `response` is emitted as a type *reference* the consumer resolves, so reading the generated text cannot distinguish `unknown` from `Widget` — a text assertion would pass either way.

Mutation-checked rather than assumed: restoring `ctx.json` in the minimal template fails the suite with `error TS18046: 'list' is of type 'unknown'` — the exact defect, caught.

- CLI suite: **588 passed** (52 files), including the existing `tsc --noEmit` runs over generated fixtures
- `pnpm docs:build` clean — validates the new `#return-value-handlers` anchor and `./typed-client.md` link
- oxlint + oxfmt + typecheck clean

## Note

Behaviour of the generated 404 changes shape: `ctx.problem.notFound()` sends RFC 9457 `problem+json` rather than the legacy `{ message }`. That is the framework's canonical error shape and what the typed client's error channel expects, but it is a visible difference for anyone diffing newly-scaffolded output against an older project. Existing generated code is untouched.

## Summary by CodeRabbit

* **New Features**
  * Generated controllers now return payloads directly, improving automatic response type inference.
  * The minimal controller pattern now includes complete CRUD endpoints.
  * Generated handlers use dedicated helpers for created and no-content responses, with clearer error handling.

* **Documentation**
  * Updated guides, examples, and generated project documentation to explain return-value handlers and response typing.

* **Bug Fixes**
  * Fixed missing or overly broad response types in scaffolded controllers.


forinda:
Valid — fixed in e3da2f53.

The emitted import was **112 characters**, over the `printWidth: 100` that `generatePrettierConfig` writes into every scaffolded project. So the adopter's first `pnpm format` would have rewritten a file the CLI had just generated. Worth noting it was already over at 105 before this branch; adding `reply` to the list made it worse rather than introducing it.

Wrapped exactly as Prettier would:

```ts
import {
  Controller,
  Get,
  Post,
  Put,
  Delete,
  Autowired,
  ApiQueryParams,
  reply,
  type Ctx,
} from '@forinda/kickjs'
```

Both call sites (DDD at L6, REST at L88) share one template literal, so a single edit covered them.

Also added a guard test asserting no generated controller exceeds 100 columns, across all three generators (minimal pattern, `g controller`, `g scaffold`). Nothing else catches this class of issue — `tsc --noEmit` passes either way, and the repo's own oxfmt only formats the template file, not the string it emits. The failure message prints the offending line numbers and lengths.

CLI suite: 589 passed.

PR #495, opened 2026-07-28
feat(db): PostgreSQL named schemas via pgSchema()

forinda:
Adds PostgreSQL named-schema support to `@forinda/kickjs-db`. Previously a table could not declare a schema at all: `TableDecl` had no field for it, the emitters never qualified an identifier, and there was no `withSchema` anywhere in the package.

## API

```ts
import { pgSchema } from '@forinda/kickjs-db/pg'
import { table, serial, varchar } from '@forinda/kickjs-db'

const billing = pgSchema('billing')

const invoices = billing.table('invoices', {
  id: serial().primaryKey(),
  ref: varchar(32).notNull(),
})

const users = table('users', { id: serial().primaryKey() }) // default search_path
```

```sql
CREATE SCHEMA IF NOT EXISTS "billing";
CREATE TABLE "billing"."invoices" ( ... );
CREATE TABLE "users" ( ... );
```

```ts
type DB = KickDbSchema // { 'billing.invoices': {...}, users: {...} }
await db.selectFrom('billing.invoices').selectAll().execute()
```

Kysely reads a dotted table name as schema-qualified, so the qualified key is what makes the query resolve with no `withSchema()` call.

## What is qualified

Everything that names the table: `CREATE` / `DROP` / `ALTER TABLE`, `CREATE INDEX`, and any `REFERENCES` pointing at it. `DROP INDEX` too — an index lives in its table's schema but `DROP INDEX` resolves the *name* through `search_path`, so a bare name would miss, or hit a same-named index in `public`. `CREATE INDEX` needs no qualifier because it takes the schema from the qualified table it targets.

Snapshot keys are qualified, so two schemas may hold same-named tables: `billing.events` and `audit.events` no longer collide.

## Three decisions worth reviewing

**`pgSchema('public')` collapses to "no schema"** — in the runtime value *and* the type key. PG puts `public` on the default `search_path`, so `pgSchema('public').table('users', …)` and `table('users', …)` name the same physical table. Keeping them as distinct snapshot keys would make the diff emit `DROP TABLE users` + `CREATE TABLE public.users` — data loss for anyone switching spellings. The conditional return type keeps the static key in step with the runtime one; a mismatch there would have every query name a table the migration never created.

**No `dropSchema` change exists.** A schema can hold objects this app never declared — another service's tables, extensions, views — so a `DROP SCHEMA` inferred from "nothing references it any more" could destroy data the diff never saw. Down migrations leave the emptied schema for an operator to remove deliberately.

**PostgreSQL only.** On MySQL a schema *is* a database and SQLite has none (only `ATTACH` aliases), so the same qualified identifiers would parse and mean something different. `extractSnapshot` throws before any DDL is written, and both emitters carry an explicit unreachable-case guard rather than silently emitting `CREATE SCHEMA` (which on MySQL means `CREATE DATABASE`).

## Backward compatibility

Unqualified tables are untouched. No `schema` field is written, no `schemas` array is added — an added-but-`undefined` key would change the serialized JSON and invalidate every existing migration hash. There is a test asserting exactly that.

## Two latent bugs this surfaced

Both were pre-existing and are fixed here, each caught by a test written for the feature rather than by inspection:

- **`diffTable` stamped the bare table name** onto every column / index / FK change instead of the snapshot key, so a qualified table emitted `ALTER TABLE "invoices"` — which resolves through `search_path` to a different table, or to nothing.
- **`invertChanges` compared `dropTable`'s bare `.name` against `dropIndex`'s qualified key**, so the redundant-teardown prune silently stopped firing for schema-qualified tables. On SQLite that prune is load-bearing, not tidiness — without it, FK/index drops compile to a table rebuild against a snapshot that no longer holds the table.

Also corrects the vitest alias for `@forinda/kickjs-db/pg`, which pointed at `src/dsl/columns/pg.ts` — narrower than the published `exports['./pg']`, so anything else on that subpath (the adapter, the dialect, and now `pgSchema`) was invisible to tests.

## Testing

- 19 behavioural tests — snapshot keying, the `public` collapse, cross-schema collision, sorted `schemas` (hash stability), name validation, the dialect guard, DDL for every statement kind, FK targets, and a full up/down round trip asserting nothing references the qualified table by its bare name.
- 5 type-level tests via `vitest --typecheck`, in a `.test-d.ts` because the package `tsconfig.json` only includes `src` — `tsc --noEmit` would never have seen them.
- **478 passed, no type errors.** The 4 failures in `introspect-sqlite.test.ts` are pre-existing and environmental: a stale `better-sqlite3` native binary (`NODE_MODULE_VERSION 127`, built for Node 22, running Node 24). They fail identically on `main`.

## Out of scope

Introspection stays single-schema — `pgAdapter({ schema })`, default `public`, which today only scopes drift-detection queries. Tables in another schema are created and migrated correctly but are **not** compared against the live database yet. Documented with a warning in `docs/api/db.md`. The runtime `withSchema()` that the parked schema-per-tenant design (`docs/db/architecture.md:536`) needs is likewise not included.

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL named-schema support through `pgSchema()`.
  * Generated migrations now create schemas when needed and correctly qualify tables, indexes, constraints, and references.
  * Tables with identical names in different schemas are handled independently, with schema-aware Kysely row-type keys.
  * The default `public` schema remains equivalent to an unqualified table.

* **Bug Fixes**
  * Corrected schema-qualified table references in column, index, and foreign-key changes.
  * Down migrations never drop schemas.

* **Documentation**
  * Added API documentation and PostgreSQL-only usage guidance.

### DIFF
--- a/.changeset/lucky-donkeys-shake.md
+++ b/.changeset/lucky-donkeys-shake.md
@@ -1,0 +1,34 @@
+---
+'@forinda/kickjs-db': minor
+---
+
+feat: PostgreSQL named schemas via `pgSchema()`
+
+Tables can now live in a named PG schema:
+
+```ts
+import { pgSchema } from '@forinda/kickjs-db/pg'
+
+const billing = pgSchema('billing')
+const invoices = billing.table('invoices', {
+  id: serial().primaryKey(),
+  ref: varchar(32).notNull(),
+})
+```
+
+- Emits `CREATE SCHEMA IF NOT EXISTS "billing"` ahead of the tables that need it.
+- Qualifies every generated statement — `CREATE`/`DROP`/`ALTER TABLE`, `CREATE INDEX`, `DROP INDEX` (whose name resolves through `search_path`), and any `REFERENCES` pointing at the table.
+- Keys the row type as `KickDbSchema['billing.invoices']`, which Kysely reads as schema-qualified, so `db.selectFrom('billing.invoices')` resolves with no `withSchema()` call.
+- Two schemas may hold same-named tables: snapshot keys are qualified, so `billing.events` and `audit.events` no longer collide.
+
+`pgSchema('public')` collapses to "no schema" in both the runtime value and the type key — PG puts `public` on the default `search_path`, so treating it as a distinct key would make the diff emit `DROP TABLE users` + `CREATE TABLE public.users` for what is the same physical table.
+
+Schemas are never dropped. There is no `dropSchema` change: a schema can hold objects this app never declared, so a `DROP SCHEMA` inferred from "nothing references it any more" could destroy data the diff never saw. Down migrations leave the emptied schema for an operator to remove.
+
+PostgreSQL only — on MySQL a schema is a database and SQLite has none, so declaring one and diffing against those dialects throws at snapshot time, before any DDL is written.
+
+Unqualified tables are completely unaffected: no `schema` field is written, no `schemas` array is added, and snapshots serialize byte-identically, so existing migration hashes stay valid.
+
+Introspection is unchanged and still single-schema (`pgAdapter({ schema })`, default `public`) — tables in other schemas are created and migrated correctly but are not yet compared during drift detection.
+
+Also fixes a latent bug this surfaced: `diffTable` stamped the bare table name onto every column/index/FK change instead of the snapshot key.
--- a/docs/api/db.md
+++ b/docs/api/db.md
@@ -263,6 +263,76 @@ See [`docs/guide/db-extensions.md`](../guide/db-extensions.md) for the full mapp
 
 PostgreSQL-only ENUM type. See [Schema Types guide](../guide/db-schema-types.md#postgresql-enums).
 
+### `pgSchema(name)`
+
+Declare a PostgreSQL named schema and hang tables off it. Imported from the
+`/pg` subpath.
+
+```ts
+import { pgSchema } from '@forinda/kickjs-db/pg'
+import { table, serial, varchar } from '@forinda/kickjs-db'
+
+const billing = pgSchema('billing')
+
+const invoices = billing.table('invoices', {
+  id: serial().primaryKey(),
+  ref: varchar(32).notNull(),
+})
+
+const users = table('users', { id: serial().primaryKey() }) // default search_path
+```
+
+Generates:
+
+```sql
+CREATE SCHEMA IF NOT EXISTS "billing";
+CREATE TABLE "billing"."invoices" ( ... );
+CREATE TABLE "users" ( ... );
+```
+
+and keys the row type by the qualified name, which Kysely reads as
+schema-qualified:
+
+```ts
+type DB = KickDbSchema // { 'billing.invoices': {...}, users: {...} }
+
+await db.selectFrom('billing.invoices').selectAll().execute()
+```
+
+Every generated statement for the table is qualified — `CREATE`/`DROP`/`ALTER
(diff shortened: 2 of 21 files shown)


forinda:
CI caught three type errors, fixed in 696a2bfc.

**Correction to the PR description above:** it claims "478 passed, no type errors." The test count was right; the type-error claim was not. `vitest --typecheck` prints `Type Errors  no errors` in its summary line *even when TypeCheckErrors are reported above it* — I grepped that summary rather than checking the exit code, so I reported the package as type-clean when it wasn't. The failures reproduce locally with plain `pnpm test`.

**1. Stale `/pg` path mapping (root cause).** `packages/db/tsconfig.test.json` mapped `@forinda/kickjs-db/pg` to `src/dsl/columns/pg.ts` — the same too-narrow mapping I fixed in `vitest.config.ts`, but for typecheck rather than runtime. I fixed one and missed the other, so tsc resolved the subpath to the column types alone and couldn't see `pgSchema`.

The four `Parameter 't' implicitly has an 'any' type` errors were a cascade from this, not separate problems: with the import unresolved, the schema object became an error type, so its `.table()` constraint callbacks lost inference. Repointing the path cleared all five.

**2. Fuzz-harness exhaustiveness.** `__tests__/fuzz/apply-changeset.ts` had no `createSchema` case, so its `const _exhaustive: never = change` guard failed — working exactly as intended. Added an applier that mirrors `extractSnapshot` (sorted, omitted when empty) so the fuzz round-trip stays an exact comparison. There's no `dropSchema` counterpart to mirror, by design.

Verified by reproducing the failure first, applying the fix, then re-running: **478 passed, exit 0, no TypeCheckErrors.** The 4 failures under a bare local `pnpm test` are the pre-existing `better-sqlite3` native-ABI mismatch on my machine (built for Node 22, running Node 24) — they passed in CI on the previous run.

**Follow-up worth a separate PR:** `packages/db-{pg,mysql,sqlite}/tsconfig.test.json` carry the identical stale mapping. I left them alone here — they're deprecated re-export shims, nothing in them imports `pgSchema`, and repointing them risks surfacing unrelated pre-existing test-type errors in those packages.

forinda:
Both addressed in 7d9e1b4a.

**Non-string schema names — valid, and worse than reported.** I checked the coercion rather than reasoning about it: `RegExp.test(x)` stringifies its argument, so the regex guard alone accepted **four** shapes, not just `undefined`:

| input | coerces to | consequence |
|---|---|---|
| `undefined` | `'undefined'` | effective schema `undefined` → silently unqualified table |
| `null` | `'null'` | `__schema = null`, and `extractTable`'s `!== undefined` check lets it through → `"schema": null` serialized into the snapshot JSON |
| `true` | `'true'` | table keyed `"true.x"`, DDL emitting `CREATE SCHEMA "true"` |
| `['a']` | `'a'` | table keyed `"a.x"`, `CREATE SCHEMA "a"` |

(`123` and `{}` were already rejected — they stringify to `'123'` / `'[object Object]'`, neither of which matches.)

The `null` case is the one I'd call a real defect rather than hygiene: it corrupts the snapshot shape, which is hashed. Added the `typeof` guard plus a test covering every case above, and confirmed the test fails without the fix — `pgSchema(undefined): expected [Function] to throw an error`.

**pnpm in the documented run command** — fixed in both `.test-d.ts` headers. I updated `columns-array-brands.test-d.ts` too, since it carried the identical line and fixing only mine would have left the pair inconsistent.

479 passed, exit 0, no TypeCheckErrors.

PR #500, opened 2026-08-05
feat(config): read .env.test under test runs so dev env stops leaking in

forinda:
## The problem

`packages/kickjs/src/config/env.ts` calls `tryLoadDotenv()` as a bare
module-top-level statement, so **any** import of `@forinda/kickjs`,
`@forinda/kickjs/config`, or `@forinda/kickjs-testing` runs:

```ts
dotenv.config({ override: false, quiet: true })   // reads ${cwd}/.env
```

`override: false` protects vars the test runner already pinned. It does nothing
for the vars it *didn't* pin — those are silently backfilled from the
developer's `.env`.

The failure mode is quiet and asymmetric: a suite can pin its database URL and
still reach live development services through every var it forgot. Nothing in
the run reports it. The test runner's `env` block ends up being the entire
defence, and it's a hand-maintained allow-list — you find out what's missing
when a test writes somewhere it shouldn't have.

Compounding it, `loadEnv()` parses `process.env` once into a module-level
`cachedEnv`. `ConfigService.get()` and `@Value()` read that snapshot, so a
`vi.stubEnv()` in `beforeAll` is a **silent no-op** for them — the parse already
happened at import time.

## The change

Under a test run (`NODE_ENV=test`, or Vitest's `VITEST`), if `.env.test` exists
in `process.cwd()`, it is read and `.env` is **not**:

```ts
function resolveEnvFiles(): string[] | null {
  const explicit = process.env['KICKJS_ENV_FILE']
  if (explicit === 'off' || explicit === 'none') return null
  if (explicit) return explicit.split(',').map((f) => f.trim()).filter(Boolean)
  if (isTestRun() && existsSync(resolve(process.cwd(), '.env.test'))) return ['.env.test']
  return ['.env']
}
```

**Deliberately not a Vite-style cascade.** A `.env.test` → `.env` fallthrough
keeps the exact hole open that this is meant to close — falling through *is* the
leak. One file wins outright.

`KICKJS_ENV_FILE` overrides the decision: a comma-separated file list, or `off`
to skip dotenv entirely (shell/Docker/secret-manager deployments).

### Two related fixes

- **`reloadEnv()`** used an inline `dotenv.config({ override: true })` hardcoded
  to `.env`. Under a test run it could clobber pinned `.env.test` values with
  the developer's own. It now routes through the same resolver as boot, so a
  reload can never disagree with startup about which file is authoritative.
- **`resetEnvCache()`** cleared `cachedEnv`/`cachedSchema` but left
  `Container._envResolver` installed — a live closure over a null cache, making
  `@Value()` reads indistinguishable from an unset key with no schema left to
  re-parse against. Now cleared.

## Backward compatibility

An app with no `.env.test` gets byte-identical behaviour, plus one stderr line
the first time a test run backfills from `.env`:

```
[kickjs] test run backfilled 14 env var(s) from .env (PORT, DATABASE_URL, …).
Create a .env.test to isolate the suite, or set KICKJS_ENV_FILE=off.
```

An opt-in nobody discovers is worth nothing, hence the warning. `KICKJS_ENV_FILE=.env`
restores the old path explicitly and silently.

Direction of failure is safe: creating a file is the safe action and takes
effect automatically; inheriting dev env into a test now requires deleting it.

## Verification

`packages/kickjs/__tests__/env-file-resolution.test.ts` covers isolation, both
`KICKJS_ENV_FILE` forms, and the no-`.env.test` fallback. It drives
`tryLoadDotenv()` through `reloadEnv()`, since the boot call only runs at module
load.

Confirmed it fails on the previous behaviour by stashing the fix:

```
× reads .env.test and does NOT fall through to .env
  AssertionError: expected 'dev' to be 'test'
× KICKJS_ENV_FILE=off skips dotenv entirely
  AssertionError: expected 'dev' to be undefined
```

The two back-compat cases pass both before and after — that's the compatibility
guarantee.

Full package suite: **707/707 pass**. `tsc --noEmit` clean, `format:check` clean.

## Docs

`docs/guide/testing.md` "Environment Isolation" is rewritten. It previously led
with `vi.stubEnv()` as the whole story and warned:

> Never use `process.env.X = 'value'` directly — it leaks across tests.

That's a wrong mental model, not just an omission — it teaches the pattern that
no-ops against `ConfigService`/`@Value()` and forbids placements that work. The
section now leads with load order, documents `.env.test` and `KICKJS_ENV_FILE`,
and shows the `resetEnvCache()` + `loadEnv()` escape hatch for mid-suite stubs.

## Not in scope

`kick new` scaffolding a `.env.test` and a `kick doctor` check for
`.env`-without-`.env.test`. Worth doing; separate PR.

## Summary by CodeRabbit

* **New Features**
  * Environment loading now follows mode-based rules, including test-specific precedence and custom file overrides.
  * Added an option to disable environment file loading entirely.
  * Added diagnostics for detecting test environments that may inherit variables from generic environment files.

* **Bug Fixes**
  * Improved environment reload behavior and test isolation.
  * Added clearer warnings and remediation guidance for missing or misconfigured test environment files.

* **Documentation**
  * Updated testing guidance with environment loading order and configuration reload recommendations.

### DIFF
--- a/.changeset/lucky-pugs-search.md
+++ b/.changeset/lucky-pugs-search.md
@@ -1,0 +1,25 @@
+---
+'@forinda/kickjs': minor
+---
+
+Read `.env.test` instead of `.env` under a test run, so development env stops leaking into tests
+
+KickJS loads dotenv as an import-time side effect with `override: false`. That
+protects vars the test runner already pinned, but every var it _didn't_ pin was
+silently backfilled from the developer's `.env` — so a suite could pin its
+database URL and still reach live development services through the vars it
+forgot, with nothing in the output saying so.
+
+Now, when `NODE_ENV=test` (or Vitest's `VITEST` is set) and a `.env.test` exists
+in `process.cwd()`, that file is read and `.env` is **not**. No cascade — falling
+through to `.env` is the leak. Set `KICKJS_ENV_FILE` to a comma-separated file
+list, or `off`, to override the choice.
+
+Apps with no `.env.test` keep their exact current behaviour, plus a one-time
+stderr warning naming what got backfilled.
+
+Also fixed: `reloadEnv()` used an inline `dotenv.config()` that always read
+`.env`, so an HMR reload could clobber pinned `.env.test` values; it now routes
+through the same resolver as boot. And `resetEnvCache()` left
+`Container._envResolver` installed over a null cache, making `@Value()` reads
+indistinguishable from an unset key — it is now cleared.
--- a/.changeset/lucky-pugs-search.md
+++ b/.changeset/lucky-pugs-search.md
@@ -10,13 +10,27 @@ silently backfilled from the developer's `.env` — so a suite could pin its
 database URL and still reach live development services through the vars it
 forgot, with nothing in the output saying so.
 
-Now, when `NODE_ENV=test` (or Vitest's `VITEST` is set) and a `.env.test` exists
-in `process.cwd()`, that file is read and `.env` is **not**. No cascade — falling
-through to `.env` is the leak. Set `KICKJS_ENV_FILE` to a comma-separated file
-list, or `off`, to override the choice.
+Env files now follow the cascade Vite popularised, where a mode-specific file
+outranks every generic one (`[mode]` is `NODE_ENV`):
 
-Apps with no `.env.test` keep their exact current behaviour, plus a one-time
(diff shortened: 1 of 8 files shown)


forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

PR #509, opened 2026-08-06
fix(core): export the types ContextDecorator resolves to

forinda:
## The bug

Every file `kick g contributor` generates fails `tsc --noEmit` in a consumer app:

```
error TS4023: Exported variable 'Tenant' has or is using name
'ContextDecoratorWithDefaults' from external module
".../node_modules/@forinda/kickjs/dist/app-module-DtoySG6l" but cannot be named.
```

`ContextDecorator` is exported, but the two interfaces the union *resolves to*
are not — `ContextDecoratorWithDefaults` and `ContextDecoratorRequiringParams`,
along with the `MissingParamKeys` / `CallSiteParams` helpers that appear in
their public positions.

`defineContextDecorator()` infers one of those concrete interfaces, so the
ordinary consumer pattern — the exact shape the generator writes —

```ts
export const Tenant = defineContextDecorator({ ... })
```

makes TypeScript emit a declaration for `Tenant` naming a type the consumer has
no import path to.

**Why it shipped:** the framework's own build stays green. Same-project types
are always nameable, so this can only ever surface downstream, in an app
consuming the built `.d.ts`.

## The fix

Re-export the four types from `core/index.ts`. The root does
`export * from './core'`, so that is the only change needed. Type-only — no
runtime behaviour moves.

## How it was found

Scaffolded an app, confirmed a clean baseline `tsc --noEmit`, then generated
contributors across the generator's options:

| Generated | Before | After |
|---|---|---|
| `g contributor tenant --type http` | TS4023 | clean |
| `g contributor clock --type bare` | TS4023 | clean |
| `g contributor project --type http --params orgId:string,limit:number` | TS4023 | clean |
| `g contributor author -m post` (module-scoped) | TS4023 | clean |

4 generated, 4 errors, 0 after. Verified by swapping the rebuilt `dist` into the
scaffolded app's `node_modules` and re-running its own `tsc --noEmit`.

## A note on the regression test

The first version of the guard was **inert** and worth calling out, because the
trap is easy to repeat.

`packages/kickjs/tsconfig.json` has `include: ["src"]`, so anything under
`__tests__` is never typechecked by `pnpm typecheck`. A test whose whole purpose
is a type-only import therefore passed happily with the export deleted.

It only becomes a real guard once registered in `test.typecheck.include` in
`vitest.config.ts`. Verified the hard way — delete the export, and it now fails:

```
TypeCheckError: Module '"../src"' has no exported member 'ContextDecoratorWithDefaults'.
 ❯ __tests__/context-decorator-public-types.test.ts:33:8
```

That allowlist already carried a comment about the same trap:

> Default include only matches `*.test-d.ts` — without this the `expectTypeOf`
> suite below was never actually type-checked.

## Verification

- **721/721** kickjs tests pass.
- `tsc --noEmit` and lint clean on the framework.
- Scaffolded app with 4 generated contributors: `tsc --noEmit` exits 0.

## Summary by CodeRabbit

* **New Features**
  * Exported additional context decorator and helper types for downstream TypeScript usage.
  * Added stronger compile-time validation for context keys across execution, request, decorator, and contributor APIs.
  * Preserved support for custom context metadata and route-specific context keys.

* **Tests**
  * Added regression coverage for public type exports, valid key usage, and invalid key detection.
  * Expanded type-checking coverage for contributor and context APIs.
  * Verified generated declarations compile successfully for downstream consumers.

### DIFF
--- a/.changeset/context-key-narrowing.md
+++ b/.changeset/context-key-narrowing.md
@@ -1,0 +1,35 @@
+---
+'@forinda/kickjs': minor
+'@forinda/kickjs-testing': patch
+---
+
+Narrow every key-taking surface to the declared `ContextMeta` / `ContextKeys` keys
+
+Augmenting `ContextMeta` narrowed `dependsOn` and nothing else. `key` on a
+contributor spec, `ctx.get`, `ctx.set`, `ctx.require` and `getRequestValue` were
+all `K extends string`, so in a fully augmented app a typo'd key still compiled
+and failed at runtime — or silently read `undefined`.
+
+The two questions are different and only one was being asked:
+
+- `TKeys` — "does **this route** carry the key?" `get` deliberately stays loose
+  here, because claiming presence wrongly fails open.
+- `ContextMetaKey` — "does this key exist in the app **at all**?" An undeclared
+  key here is a typo, not a maybe-absent value.
+
+All key positions now use `ContextMetaKey`. Value types were already correct and
+are unchanged: `ctx.get('user')` is `User | undefined`, `ctx.require('user')` is
+`User`, and a key registered only in `ContextKeys` still reads as `unknown`.
+
+`ContextMetaKey` moved next to the registries it reads in `execution-context`,
+so that module can constrain `get`/`set` without a circular import. It is still
+re-exported from `context-decorator`, so existing imports are unaffected.
+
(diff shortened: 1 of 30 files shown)


forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

PR #511, opened 2026-08-07
feat(spa): engine-agnostic declarative SpaAdapter, three fallback bugs fixed

forinda:
## The headline: the documented usage crashed

`docs/guide/spa.md` has always shown the factory form:

```ts
SpaAdapter({ clientDir: 'dist/client', apiPrefix: '/api' })
```

The export was a `class`. So following the guide threw:

```text
TypeError: Class constructor SpaAdapter cannot be invoked without 'new'
```

`ViewAdapter` moved to `defineAdapter()` in v4 and its docblock records it —
SpaAdapter was simply missed. It is now a factory too, so the code matches the
docs and its sibling.

## It only ever worked on Express

The adapter reached for `express.static`, `req.path`, and `res.send`. None of
those exist under Fastify or h3, so on those runtimes it silently served
nothing — while `bootstrap({ runtime })` advertises a pluggable engine and
`AdapterContext` documents `http` as *"the supported way… an adapter written
against `http` works under any runtime."*

Rewritten against `http.serveStatic()` / `http.use()`, reading the path from
`req.url` and writing through `res.setHeader` / `res.end`.

## Three fallback bugs

Confirmed by probing a real Express app **before** changing anything:

| Request | Before | After |
| --- | --- | --- |
| `GET /users/john.doe` | 404 | `index.html` |
| `HEAD /dashboard` | 404 | 200, headers only |
| `GET /apidocs` (`apiPrefix: '/api'`) | 404 | `index.html` |

- **Dotted routes** — `req.path.includes('.')` treated every dotted path as a
  file request. Replaced with content negotiation: fall back for `GET`/`HEAD`
  that accept HTML. A missing `/assets/app.js` still 404s rather than receiving
  an HTML document the browser cannot parse as JavaScript. `alwaysFallback: true`
  opts out for non-browser clients.
- **HEAD** — dropped by a `!== 'GET'` check, breaking proxies and link checkers.
  Now returns identical headers with no body.
- **Prefix boundary** — `startsWith('/api')` swallowed `/apidocs`. Matching is
  segment-aware: `/api` covers `/api` and `/api/users` only.

## A regression I introduced, and caught

Moving cache headers out of `express.static`'s `setHeaders` — which only fires
on a **hit** — into standalone middleware meant an unconditional header. The
end-to-end run showed it:

```text
GET /assets/missing.js   404  cc=public, max-age=31536000, immutable
```

A year-long immutable cache on a 404. Headers are now applied only to paths
resolving to a real file inside `clientDir`, with a containment check so the
stat cannot be walked out of the directory:

```text
GET /assets/missing.js   404  cc=-
```

## Verification

The adapter had **no tests**, despite being public API with its own export path
(`@forinda/kickjs/spa`). Adds 21, driving it through a recording `AdapterHttp`
stub and replaying the middleware against plain node-shaped objects — so
anything Express-specific fails here rather than in a consumer's Fastify app.

Every fix was verified by re-introducing the old rule and watching the matching
test fail:

```text
old dot heuristic   → × serves a route containing a dot (was 404)
old GET-only rule   → × answers HEAD with headers and no body (was 404)
old startsWith      → × does not treat /apidocs as an API route (was 404)
                      × skips excluded paths, segment-aware
removed containment → × refuses to stat outside clientDir
```

Plus an end-to-end pass against a real Express app: SPA routes, dotted routes,
HEAD, prefix lookalikes, a real API route, a static asset, and a missing asset
all behave correctly.

**747/747** tests, `pnpm typecheck` clean, lint clean.

## Breaking

`new SpaAdapter({...})` → `SpaAdapter({...})`. Anyone following the guide was
already writing the new form (and hitting the TypeError); only callers who
worked around the docs with `new` need the edit.

## Summary by CodeRabbit

* **New Features**
  * SPA serving now works consistently across supported HTTP engines.
  * Added configurable `alwaysFallback` behavior and improved handling for dotted paths, `HEAD` requests, API prefixes, exclusions, and HTML content negotiation.
  * Static assets receive cache headers only when successfully found.
  * Fullstack project scaffolding now supports production frontend serving, root start commands, and safe client-directory configuration.

* **Documentation**
  * Expanded SPA integration guidance covering compatibility, fallback rules, routing boundaries, and configuration options.

* **Tests**
  * Added comprehensive coverage for serving, fallback behavior, caching, missing builds, path safety, and generated fullstack setups.

### DIFF
--- a/.changeset/fullstack-spa-serving.md
+++ b/.changeset/fullstack-spa-serving.md
@@ -1,0 +1,26 @@
+---
+'@forinda/kickjs-cli': minor
+---
+
+Fullstack template: serve the built frontend from the API origin
+
+`kick new --template fullstack` scaffolds `server/` + `web/` and wires a typed
+dev loop — Vite serves the client and proxies `/api` to the server. But it
+stopped at the deploy boundary: `pnpm build` produced `web/dist` and **nothing
+ever served it**. The generated bootstrap had no adapters, and the root had no
+`start` script — only `vite preview`, which is a dev preview server.
+
+The generated server now wires `SpaAdapter({ clientDir: '../web/dist' })`, and
+the root gains a `start` script that runs the server. So:
+
+|            | dev                       | production               |
+| ---------- | ------------------------- | ------------------------ |
+| `web/dist` | absent                    | present                  |
+| SpaAdapter | inert (registers nothing) | serves `/`               |
+| `/api/v1`  | Vite proxy → `:3000`      | controllers, same origin |
+
+Dev is unchanged: the adapter early-returns while the build directory does not
+exist, which is the normal state under `kick dev`.
+
+`generateEntryFile` takes an optional `spaClientDir` and composes it with the
+existing swagger / devtools adapter injection rather than replacing it.
--- a/.changeset/spa-adapter-engine-neutral.md
+++ b/.changeset/spa-adapter-engine-neutral.md
@@ -1,0 +1,38 @@
+---
+'@forinda/kickjs': major
+---
+
+SpaAdapter: engine-agnostic, declarative, and three fallback bugs fixed
+
+**The documented usage did not work.** `docs/guide/spa.md` has always shown
+`SpaAdapter({ ... })`, but the export was a class — so following the guide threw
+`TypeError: Class constructor SpaAdapter cannot be invoked without 'new'`. It is
+now a `defineAdapter()` factory, matching both the docs and `ViewAdapter`, which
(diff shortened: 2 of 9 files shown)


forinda:
@coderabbitai full review

forinda:
@coderabbitai  full review

forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

PR #518, opened 2026-08-07
feat(http): add ctx.ip and ctx.redirect — engine-neutral by construction

forinda:
Closes the two API gaps the generator audit surfaced. The guide reached for
`ctx.req.ip` and `ctx.res.redirect()` because nothing neutral existed — both are
engine-native:

| API | Express | Fastify | h3 |
| --- | --- | --- | --- |
| `req.ip` | ✓ | ✓ | ✗ |
| `res.redirect()` | ✓ | ✓ | ✗ |

So the docs had to carry "this breaks on X" caveats instead of showing portable
code. Those caveats are deleted here.

## `ctx.ip`

Prefers the address **the runtime computed** — Express derives `req.ip` from
`trust proxy`, Fastify from `trustProxy` — because raw forwarded headers are
client-**spoofable** wherever nothing normalizes them. Letting a header win would
let a caller choose their own rate-limit bucket or audit identity.

Falls back to `cf-connecting-ip` / `x-forwarded-for` / `x-real-ip` only for
runtimes that compute no address (the web/edge entry), then the node socket.

That resolution already existed inside `rate-limit-guard`, comment and all. It
moves onto the context and the guard reuses it, rather than a second copy
drifting from the first.

## `ctx.redirect(url, status = 302)`

Writes the status and `Location` header through the runtime response surface, so
it works everywhere. The docblock warns against an unvalidated user-supplied
URL — an attacker-controlled destination is an open redirect, a standard
phishing primitive.

## A test that was passing for the wrong reason

`edge-stores.test.ts` faked the context with an object literal cast `as never`.
`ctx.ip` is a class getter, so a literal has no such property and the rate-limit
key silently became `'global'` — the assertion broke the moment `defaultKey`
started reading it.

Before writing that off as a test artifact I checked whether the edge entry
really builds a `RequestContext`: `src/http/web/handler.ts:139` does. So the
getter is present in production and only the fake was wrong. The test now
constructs a real context, which is what makes the assertion mean anything.

## Verification

8 new tests for the accessors, including the spoofing case. Verified by deleting
the runtime-ip preference and watching **3** tests fail across both suites:

```text
× prefers the address the runtime computed
× does not let a forwarded header override a computed address
× default key prefers runtime-computed req.ip over spoofable headers
```

Docs updated: `authorization.md`, `authentication.md`, `context-decorators.md`
now use the neutral accessors, and `ctx.redirect` is in the response-helpers
table.

**787/787**, `pnpm typecheck` and `vitest --typecheck` both exit 0, lint clean.

## Summary by CodeRabbit

- **New Features**
  - Added runtime-neutral `ctx.ip` support with socket and proxy-aware fallbacks.
  - Added `ctx.redirect(url, status?)` with custom status codes across runtimes.
  - Improved rate limiting, CSRF exclusions, and request logging outside Express.
- **Bug Fixes**
  - Fixed path detection, client identification, redirect handling, and middleware behavior across supported runtimes.
  - Clarified and strengthened cross-runtime error-handler contracts.
- **Documentation**
  - Updated authentication, authorization, controller, decorator, and middleware guidance.
- **Tests**
  - Added coverage for IP resolution, redirects, middleware compatibility, logging, and error handling.

### DIFF
--- a/.changeset/ctx-ip-redirect.md
+++ b/.changeset/ctx-ip-redirect.md
@@ -12,8 +12,9 @@ carry "this breaks on X" caveats instead of showing portable code.
 **`ctx.ip`** resolves `req.ip` → `socket.remoteAddress` → forwarded headers. It prefers the address the runtime computed — Express derives `req.ip`
 from `trust proxy`, Fastify from `trustProxy` — because raw forwarded headers
 are client-**spoofable** on deployments that do not normalize them. It falls back
-to `cf-connecting-ip` / `x-forwarded-for` / `x-real-ip` only for runtimes that
-compute no address (notably the web/edge entry), then to the node socket.
+to the node socket address next, and consults `cf-connecting-ip` /
+`x-forwarded-for` / `x-real-ip` only when neither exists (notably the web/edge
+entry) — a spoofable header must never outrank a real connection address.
 
 That resolution already existed inside the rate-limit guard; it now lives on the
 context and the guard reuses it rather than keeping a second copy.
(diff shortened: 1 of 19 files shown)


forinda:
@coderabbitai full review

PR #519, opened 2026-08-07
fix(http): tear down the previous app on HMR rebuild

forinda:
Framework fix for a leak diagnosed in the field: one API process holding several
Kafka consumer-group members, a crashed `handleUpgrade()`, and migration jobs
stuck unprocessed — all one bug.

## Root cause

`bootstrap()` keeps the live app on `globalThis.__app`. The reload path already
tested for it, then replaced it without tearing it down:

```ts
if (g.__app) {
  const freshApp = new Application(options)
  g.__app = freshApp        // old one dropped, still running
}
```

`shutdown()` ran only from the `SIGINT` / `SIGTERM` handler. So every save in a
dev session stacked another live adapter set on the last.

That matches the symptoms exactly. On a single-partition topic only one group
member can hold the assignment, and it was a leaked consumer from an earlier
reload wired to nothing — so jobs silently stopped being processed and the group
had never committed an offset. A second Socket.IO server on the same HTTP server
crashed `handleUpgrade()` by the same mechanism.

## Why the obvious fix breaks dev

`shutdown()` step 1 is `httpServer.close()`, and in dev the server is **shared
across rebuilds** via `globalThis.__kickjs_httpServer`. Calling plain
`shutdown()` on reload closes the listening socket and kills HMR on the first
save — almost certainly why the call was omitted rather than forgotten.

So `shutdown()` gains `{ closeServer?: boolean }` and the reload path passes
`false`:

| Step | SIGINT | HMR reload |
| --- | --- | --- |
| close HTTP server | ✓ | — shared across rebuilds |
| drain in-flight | ✓ | — server keeps serving |
| plugin + adapter `shutdown()` | ✓ | ✓ |
| release framework resources | ✓ | ✓ |

Teardown failures are logged, not thrown — one broken adapter must not leave the
dev server with no app at all.

## Verification

8 tests. The important one drives `bootstrap()` **twice**, which is exactly what
a save does, and asserts one teardown plus a still-listening server. Verified by
removing the call and watching it fail:

```text
× shuts the previous app down before replacing it
× tears plugins down too, not only adapters   (with step 3's plugin line removed)
```

Checked against the reporting app: its Kafka adapter calls
`consumer.disconnect()` and its Socket.IO adapter calls `io.close()`. Both were
already correct — they were simply never invoked.

## Docs, including a correction

`hmr.md` listed database pools, Redis clients, and the Socket.IO server as
**preserved across HMR**. They were not preserved so much as *abandoned* — that
was the bug. Those rows now say "rebuilt", with the reason.

The consequence most likely to bite adapter authors is documented too:
`shutdown()` now runs on every rebuild and the adapter is mounted again, often as
the **same instance** (`config/adapters.ts` may not re-evaluate). An adapter must
release its handle *and clear the reference* — one that closes a resource without
nulling it works on first boot and fails on the first save.

The trade is deliberate: reconnecting a pool per reload costs a moment, leaking
one costs a debugging session.

**787/787**, typecheck and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Improved hot reload behavior by cleaning up the previous application before rebuilding.
  * Preserves the development server and active connections during hot reloads.
  * Recreates adapters, plugins, database, Redis, and Socket.IO resources cleanly.

* **Bug Fixes**
  * Prevents resource leaks during hot reloads.
  * Teardown errors are logged while allowing the rebuild to continue.
  * Ensures one failed cleanup does not block other shutdown operations.

* **Documentation**
  * Updated hot reload and application lifecycle guidance, including restartable adapter requirements.

### DIFF
--- a/.changeset/hmr-adapter-teardown.md
+++ b/.changeset/hmr-adapter-teardown.md
@@ -33,3 +33,12 @@ is skipped too — the server keeps serving, so there is nothing to drain toward
 
 Teardown failures are logged rather than thrown, so one broken adapter cannot
 leave the dev server with no app at all.
+
+Plugins are torn down alongside adapters — `shutdown()` step 3 always ran both,
+so a plugin holding a timer or connection leaked identically.
+
+Docs: the HMR guide previously listed database pools, Redis clients, and the
+Socket.IO server as _preserved across HMR_. They were not preserved so much as
+abandoned, which is the bug. They are now rebuilt, and the guide says so, along
+with what an adapter must do to be restartable — release the handle **and clear
+the reference**, since the same instance is often mounted again.
--- a/.changeset/hmr-adapter-teardown.md
+++ b/.changeset/hmr-adapter-teardown.md
@@ -1,0 +1,35 @@
+---
+'@forinda/kickjs': patch
+---
+
+Dispose the previous app on HMR rebuild instead of leaking it
+
+`bootstrap()` keeps the live app on `globalThis.__app`, and the reload path
+already tested for it — then replaced it without tearing it down:
+
+```ts
+if (g.__app) {
+  const freshApp = new Application(options)
+  g.__app = freshApp // old one dropped, still running
+}
+```
+
+`shutdown()` ran only from the `SIGINT` / `SIGTERM` handler, so in a dev session
+every save added another live adapter set on top of the last.
+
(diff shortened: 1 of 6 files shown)


forinda:
@coderabbitai full review

PR #537, opened 2026-08-26
feat(http): opt out of URL versioning with defaultVersion: false

forinda:
## The gap

`/api/v{n}` was mandatory. `apiPrefix` could be changed or emptied, but nothing removed the version segment:

```ts
bootstrap({ modules })                    // /api/v1/todos
bootstrap({ modules, apiPrefix: '' })     // /v1/todos   ← still versioned
```

An API that doesn't version its URLs had no way to say so.

## The change

`defaultVersion` now accepts `false`:

```ts
bootstrap({ modules, defaultVersion: false })                    // /api/todos
bootstrap({ modules, apiPrefix: '', defaultVersion: false })     // /todos
```

`ModuleRoutes.version` accepts `false` too, so the **per-mount value keeps overriding the app default in both directions** — that's the escape hatch:

```ts
bootstrap({ modules, defaultVersion: false })
routes: () => [
  { path: '/todos', controller: TodoController },                    // /api/todos
  { path: '/payments', version: 2, controller: PayController },      // /api/v2/payments
]

bootstrap({ modules })  // defaultVersion: 1
routes: () => [
  { path: '/todos', controller: TodoController },                    // /api/v1/todos
  { path: '/webhooks', version: false, controller: HookController }, // /api/webhooks
]
```

The second direction is the one that earns its keep: a webhook receiver or health surface a third party already hardcoded, living inside an otherwise versioned app.

Additive. The default stays `1`, and every existing app mounts exactly where it did.

## Why a shared helper instead of two ternaries

The mount path was interpolated in **two** places — `Application` (node) and `createWebApp` (web/edge):

```
packages/kickjs/src/http/application.ts:778
packages/kickjs/src/web.ts:184
```

Both now call `buildMountPath()` in `core/path.ts`. Patching only the node site — the one the request names — would have shipped an edge entry that silently kept versioning, and the duplicate expressions would drift again on the next change.

Routing through `joinPaths` also fixed two things the raw template got wrong:

- **Floors the result at `/`.** Express throws on `app.use('', router)`, which `apiPrefix: '' + defaultVersion: false` produces directly. Covered by a test.
- **Collapses a trailing slash on the prefix.** `apiPrefix: '/api/'` used to mount at `/api//v1/todos`.

## Verification

12 new tests in `packages/kickjs/__tests__/api-versioning-optout.test.ts` — `buildMountPath` units plus HTTP round-trips through both entries: default mount, opt-out, the versioned path 404ing after opt-out, root mount, per-module override in each direction, and `createWebApp`.

Sabotage-checked: dropping the `version === false` branch from the helper fails 7 of the 12, including the web-entry test — so the shared-helper claim is exercised, not assumed.

`842 kickjs tests pass`, full monorepo `44/44 turbo test tasks`, `27/27 typecheck`, `pnpm docs:build` clean.

## Docs

- `docs/guide/modules.md` — new "Opting out of versioning" section, with a warning that flipping it moves every URL and is a breaking change to your own consumers (the typed client's `baseUrl` has to move with it)
- `docs/api/core.md`, `docs/api/http.md` — option tables

## Summary by CodeRabbit

* **New Features**
  * Added optional API URL versioning, allowing applications and individual modules to omit version segments.
  * Added support for mounting APIs at the root path with an empty API prefix.
  * Added per-module version overrides, including numeric versions or no version segment.
  * Added normalized API mount-path construction support across application entry points.

* **Documentation**
  * Updated API and module guides with versioning, root-mounting, overrides, and URL compatibility details.


forinda:
@coderabbitai check full review

PR #553, opened 2026-08-30
feat(cli): remove kick g job, gate optional-package imports, document the extension API

forinda:
Removes `kick g job`, gates optional-package imports in the generators that stay, and documents the extension API properly.

Every other generator is retained.

## The line: is the interface ours?

Generators for interfaces **KickJS defines** stay — `plugin`, `middleware`, `guard`, `contributor`, `adapter`, `dto`, `controller`, `service`, `module`, `scaffold`. Each scaffolds a shape the framework owns: an `AppAdapter` with the right lifecycle hooks, a `KickPlugin` with the right registration fields, a contributor with its `ContextMeta` stub, a DTO wired to the configured schema library. Those are easy to get subtly wrong by hand, and wrong fails at boot rather than at the keyboard. Maintaining a generator for a shape we define is work the framework owes adopters.

`kick g job` was not that. Its shape belonged to whichever queue the project actually runs — something the framework cannot know and should not track — and it imported `@forinda/kickjs-queue`, which no template installs, so in most projects it wrote a file that could not compile.

## Generated code that could not compile

The same defect ran quietly through `kick g module` / `kick g scaffold`: every generated REST controller carried `import { ApiTags } from '@forinda/kickjs-swagger'` and five decorators, whether or not the project depended on swagger. Nobody asked for swagger, and the file arrived broken.

Now gated on the project declaring the dependency — read from `package.json`, not resolved from `node_modules`, because what a project declares is what its generated code may import.

Two controller generators had the hardcoded import and `kick g scaffold` a third call site, so fixing the one first found would have left the others. `kick new` was already correct: its swagger and devtools imports are gated on `--packages`, which also installs them.

The four generators that were briefly removed and then restored were audited rather than assumed: they emit imports only from `@forinda/kickjs` (9) and `vitest` (1), both present in every scaffold. Nothing to gate.

## A test fixture that could not fail

The fixture symlinked `@forinda/kickjs-swagger` into every generated project, commented *"some scaffolds also import from it"* — which is exactly why these tests never caught the bug. The generated controller imported the package regardless, the symlink made it resolve, `tsc --noEmit` passed, and every real adopter got `TS2307`. The test had been fixed instead of the generator.

The symlink is gone and everything still compiles, which is the evidence the gate works.

Stated plainly: that compile check still does not cover the module templates' optional imports — sabotaging the gate fails no test today. It deserves its own change; it is the check that would have caught this class of bug without anyone reporting it.

## A documented command that never existed

`generators/auth-scaffold.ts` had no importers and no registration, while `docs/guide/generators.md` documented `kick g auth-scaffold` in full. Both removed.

## Extension docs

`docs/guide/plugin-generators.md` grows from 104 to ~340 lines: the full `GeneratorSpec` / `GeneratorContext` / `GeneratorFile` reference, a worked `job` generator with flags and defaults, a testing section (`files()` is a pure function, callable without the CLI), dispatch and conflict rules, and the ownership rule above.

Writing it surfaced two documentation bugs:

- **`projectRoot` was undocumented entirely**, even though paths resolve against `ctx.cwd` — which is wherever the adopter ran the command. A generator writing `${ctx.modulesDir}/…` from inside `src/modules/orders/` lands in `src/modules/orders/src/modules/…`.
- **`args` / `flags` are descriptors, not validation**, despite `required: true` implying otherwise.

## Verification

687 tests across 64 files. Lint, token lint, format, `tsc --noEmit`, `docs:build` all clean.

## Summary by CodeRabbit

* **New Features**
  * Generated modules and scaffolds now include Swagger annotations only when Swagger is declared in the project.
  * Repository configuration accepts `inmemory` or any custom repository name.

* **Breaking Changes**
  * Removed the built-in `kick g job` generator. Create queue-specific generators instead.
  * Prisma and Drizzle repository presets are no longer provided; they now generate a generic repository stub.

* **Documentation**
  * Expanded plugin generator guidance with custom generator APIs, testing, dispatch rules, and queue-job examples.
  * Updated CLI examples and configuration references.


forinda:
@coderabbitai full review

PR #558, opened 2026-08-30
fix(cli): terminate recursive response types instead of expanding them forever

forinda:
## What this actually was

Diagnosed — by me, and independently in the consuming app — as superlinear memory scaling in the client route map, needing >12 GB and blocking a migration. It was neither superlinear nor scaling. **Two individual routes.**

A type alias to an object literal (`type Node = { … }`) carries TypeScript's anonymous `__type` symbol rather than the alias name, so `isNamed` returned false, the type was never hoisted, and it was re-expanded inline at every occurrence. For a **recursive** alias — zod v4's `JSONSchema` (`$defs?: Record<string, JSONSchema>`), or a condition tree — inlining cannot terminate, so the walk fanned out exponentially.

An `interface` in the same position was always fine: it hoists, the hoist keys the memo, and the memo ends the recursion. Only the alias form was affected, which is why this went unnoticed.

## Measured, on a 1,940-route app

| | before | after |
| --- | --- | --- |
| whole pass | 130 s, 15.5 GB, **V8 abort, no map** | 13.4 s, 1.66 GB, map emitted |
| depth-guard warnings | 325 (8.1 M on the worst route) | **0** |
| longest emitted line | ~7,000 chars | 1,644 |
| routes / typed responses / bare `any` | 1940 / 1195 / 7 | 1940 / 1195 / 7 |

Fidelity is identical — same routes, same typed-response count, same `any` count. Only the pathological expansion is gone.

## What it emits now

The recursion is represented exactly, as mutually-referencing interfaces, instead of a duplicated blob that bottomed out in `unknown` anyway:

```ts
interface __T638 { readonly all: readonly (__T638 | __T639 | __T640 | { readonly p: string; readonly v: unknown })[] }
interface __T639 { readonly any: readonly (__T638 | __T639 | __T640 | { readonly p: string; readonly v: unknown })[] }
interface __T640 { readonly not: __T638 | __T639 | __T640 | { readonly p: string; readonly v: unknown } }
```

An anonymous object type that reaches itself mid-render claims a name and is promoted to a hoisted interface; the body is then re-rendered from depth 0, which also recovers whatever the first inline pass truncated on the way in. Non-recursive anonymous types are untouched and still render inline — there's a test pinning that.

## Backstop

Past 1,000 depth-guard hits in one route, the expansion is abandoned, emits `unknown`, and says so **once**.

The guard was firing correctly on every one of those 8.1 M warnings. It just never stopped, so the cost was unbounded — a correct guard with no bail is still an OOM. Diagnosing it meant counting warnings to notice one route was responsible.

## Testing

`packages/cli/__tests__/client-recursive-types.test.ts` drives `resolveClientMap` over a real project. The existing expander harness does **not** reproduce this — it reads the type off its own alias declaration, where it still renders by name, so tests written there passed with the fix reverted and were worthless. All five assertions in the new file fail with the fix reverted.

Also two budget tests (`truncationBudget` is injectable so they're load-bearing). Full suite green: 732 CLI tests, 2,254 across the monorepo.

## Note

Chunked resolution was prototyped against the wrong diagnosis and is **not** in this PR — with the actual cause fixed, 1.66 GB needs no chunking.

---

## Follow-up commit: `typegen.client: { maxDepth }`

The depth limit was the one ceiling in the map an adopter could not lift — hardcoded at 12, no config path. Now:

```ts
typegen: { client: { maxDepth: 24 } }
```

The object form is on unless it says otherwise, so `{ maxDepth: 24 }` alone also enables the map; `{ enabled: false }` disables it. `client: true` is unchanged.

Depth joins the map's fingerprint — without that, raising it would serve the cached map built at the old depth. Verified by flipping to 6: cache misses, 41 truncation warnings, smaller map.

**The default is not wrong, and this isn't an admission that it is.** Recursive and named types both hoist into their own interface, costing one level rather than the whole budget, so only deep *anonymous* nesting spends it. On the 1,940-route app the emitted map is **byte-identical at 12, 24, 48 and 96** — measured cold each time. The knob is for the project where that isn't true.

Changesets: `patch` for the fix, `minor` for this.

## Summary by CodeRabbit

- **New Features**
  - Configure client response-type expansion depth with `typegen.client.maxDepth`.
  - Support object-based client configuration, including explicit enable/disable controls.
  - Preserve recursive response types by generating reusable named interfaces.
  - Safely fall back to `unknown` with a single warning when expansion limits are exceeded.
  - Configuration changes now trigger client-map regeneration instead of using stale cached output.

- **Documentation**
  - Added guidance for configuring expansion depth and understanding generated recursive types.

### DIFF
--- a/.changeset/curly-moons-repeat.md
+++ b/.changeset/curly-moons-repeat.md
@@ -1,0 +1,40 @@
+---
+'@forinda/kickjs-cli': patch
+---
+
+Fix unbounded expansion of recursive response types in the client route map.
+
+A type alias to an object literal (`type Node = { … }`) carries TypeScript's
+anonymous `__type` symbol rather than the alias name, so the expander did not
+consider it nameable and re-expanded it inline at every occurrence. For a
+_recursive_ alias — zod v4's `JSONSchema`, whose `$defs` is
+`Record<string, JSONSchema>`, or any condition tree — inlining cannot
+terminate, and the walk fanned out exponentially.
+
+Measured on a 1,940-route app: one route produced 8.1 million depth-guard
+warnings and exhausted a 24 GB heap without emitting a map. A second route
+emitted a single 7,000-character property that duplicated its subtree at every
+level and still bottomed out in `unknown`. Neither is a scaling problem — both
+came from individual routes.
+
+An anonymous object type that reaches itself while being rendered now claims a
+name and hoists, so the cycle terminates on that name. Recursive types are
+emitted exactly, as mutually-referencing interfaces, instead of degrading:
+
+```ts
+interface __T638 { readonly all: readonly (__T638 | __T639 | __T640 | …)[] }
+interface __T639 { readonly any: readonly (__T638 | __T639 | __T640 | …)[] }
+interface __T640 { readonly not: __T638 | __T639 | __T640 | … }
+```
+
+On that same app: 130s / 15.5 GB / V8 abort with no output → 13.4s / 1.66 GB,
+all 1,940 routes emitted, 325 depth-guard warnings → 0, longest line 7,000 →
+1,644 characters. Route count, typed-response count and `any` count are
+unchanged, so the fidelity is identical — only the pathological expansion is
+gone.
+
+Expansion is now also bounded as a backstop: a route that hits the depth guard
+more than 1,000 times abandons its expansion, emits `unknown`, and says so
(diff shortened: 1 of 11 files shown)


forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

PR #588, opened 2026-08-31
fix(kickjs): make csrf, rateLimit and session work on Fastify and h3

forinda:
The sweep. It found more than I expected: **three shipped middlewares are unusable on Fastify and h3.**

## What the matrix found

Express passes everything. Fastify and h3 fail 8 of 27 cases:

| middleware | Express-only call | effect on Fastify / h3 |
| --- | --- | --- |
| `csrf` | `res.cookie()` | threw **before any check ran**, so every request became a **500** — including the safe methods it exists to wave through |
| `rateLimit` | `res.status().json()` | threw once the limit was hit, and the throw left the request **hanging** rather than answering 429 |
| `session` | `res.cookie()` | threw on every response issuing a session, so the cookie was never set and **every visitor looked new** |

Connect middleware receives an Express response under Express and a raw `ServerResponse` under Fastify and h3. These three reached for conveniences that only exist on the former.

`csrf` fails loudly — an app mounting it on Fastify serves 500 for everything, which you would notice in a minute. **`session` fails quietly**, which is worse: no cookie, no error, every request a new session.

## Fix

`setCookie` / `sendJson` helpers that take the Express path when it exists — so Express behaviour is byte-identical — and fall back to `Set-Cookie` and the Node response primitives otherwise. Cookie serialisation matches Express's semantics, including `maxAge` in milliseconds and `sameSite: true` → `Strict`.

## Why this survived

The existing `middleware-engine-neutral.test.ts` tests the **helpers** — `resolveClientIp`, `resolvePathname` — in isolation. It never mounts a middleware on an engine. So the unit tests were green while the middleware was unusable on two of three supported runtimes.

Its own header even documents the class of bug (`rateLimit` keying every caller as `127.0.0.1` because `req.ip` is Express-only). The response side had the same problem and no equivalent coverage.

## Matrix

`cors`, `helmet`, `requestId`, `session`, `rateLimit`, `csrf` × Express, Fastify, h3 — 30 cases, driven end to end through `createTestApp`.

`cors`, `helmet` and `requestId` were already consistent, which is worth recording: it bounds the problem to middlewares that **write** a response or a cookie, not those that only set headers.

Reverting the three call sites fails 13 cases.

Two of my own test bugs on the way, both caught by Express failing alongside the others — a plain `request()` instead of a cookie-persisting `request.agent()`, and writing to `ctx.session` rather than `ctx.session.data`. When all three runtimes fail identically, it is the test.

Monorepo green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CSRF, rate-limiting, and session middleware failures on Fastify and h3 runtimes.
  * Middleware responses and cookies now work consistently across supported runtimes.
  * CSRF ignored paths are now handled correctly across request environments.

* **Tests**
  * Added end-to-end coverage for middleware behavior across Express, Fastify, and h3.

### DIFF
--- a/.changeset/quiet-bats-repeat.md
+++ b/.changeset/quiet-bats-repeat.md
@@ -1,0 +1,28 @@
+---
+'@forinda/kickjs': patch
+---
+
+Fix `csrf`, `rateLimit` and `session` breaking on Fastify and h3.
+
+Connect middleware receives an Express response under Express and a raw
+`ServerResponse` under Fastify and h3. Three shipped middlewares reached for
+Express-only conveniences on it:
+
+| middleware  | call                  | effect on Fastify / h3                                                                                                 |
+| ----------- | --------------------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `csrf`      | `res.cookie()`        | threw before any check ran, so **every request became a 500** — including the safe methods it is meant to wave through |
+| `rateLimit` | `res.status().json()` | threw once the limit was hit, and the throw left the request **hanging** instead of answering 429                      |
+| `session`   | `res.cookie()`        | threw on every response issuing a session, so no cookie was ever set                                                   |
+
+`csrf` is the worst of the three: an app that mounted it on Fastify or h3 served
+500 for everything, which is a hard failure rather than a quiet one — but
+`session` failed quietly, and every visitor looked new.
+
+All three now go through `setCookie` / `sendJson` helpers that take the Express
+path when it exists — so behaviour under Express is unchanged — and fall back to
+`Set-Cookie` and the Node response primitives otherwise.
+
+Found by running the middleware suites as a runtime matrix rather than against
+Express alone. The existing "engine-neutral" tests check the helpers
+(`resolveClientIp`, `resolvePathname`) in isolation; they never mount a
+middleware on an engine, which is why this survived.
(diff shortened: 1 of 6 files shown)


forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

forinda:
Confirmed and fixed in 647e077a — and it is worse than described: **the double-submit flow could never succeed on ANY runtime, Express included.**

Reproduced with a round trip — take the issued cookie, send it back with the matching header:

```
express → 403
fastify → 403
h3      → 403
```

Express fails for the same reason: `req.cookies` is populated by `cookie-parser`, which an app has to mount. Without it Express sees no cookie either, so the middleware minted a fresh token on every request and compared the submitted header against that new value.

Cookies now come from a shared `readCookies` that falls back to parsing the `Cookie` header. As you point out, `session` already had exactly that fallback — it is now moved into the shared module and reused, rather than copied, so the two cannot drift apart.

**Why my matrix missed it, since that is the more useful lesson:** both csrf cases expected a *rejection*. `no token → 403` passes whether the middleware works or is completely broken — a test that cannot fail in the direction that matters. Only the round trip separates them, and that is the case I had not written.

Added exactly that, per your suggestion. Reverting the cookie read fails it on all three runtimes.

PR #604, opened 2026-08-31
docs: fix and shorten the READMEs

forinda:
## Wrong — fixed regardless of length

| file | was | why it matters |
| --- | --- | --- |
| `packages/testing` | its **only** example destructured `expressApp` and drove it with supertest | that property is deprecated and **throws** under Fastify and h3 — the package's headline example was broken on two of three supported runtimes |
| root README | `kick new` template list offering `ddd \| cqrs` | neither exists; `fullstack` was missing, though the same README demos it 200 lines earlier |
| root README | `kick g module --repo prisma` / `--repo drizzle` | now produces a generic stub, not a Prisma/Drizzle repository |
| root README | `kick g job` | hard error — queue processors are no longer built in |
| root README | `kick add swagger drizzle` (×2) | `drizzle` is not an installable name |
| `packages/mcp` | `@Roles('admin')`, `@Public()` | both from the deleted auth package |
| `packages/mcp` | "Express pipeline" ×5 | the adapter dispatches through the shared HTTP pipeline on any runtime |
| `packages/cli` | `--packages auth,…`, `--template rest\|minimal` | `auth` is dead; `fullstack` missing; generator list omitted plugin/contributor/test/agents |
| `packages/devtools-kit` | `>=5.0.0` / `^5.0.0` | package is at 7.0.1 |

Also fixed the CLI's **own** `--packages` help string, which had the same `auth` example — changeset included.

**The health module was documented nowhere.** The root README went further and taught hand-writing `@Get('/health')` in its Hello World — the exact duplication the module exists to remove. Both it and `packages/kickjs` now describe the built-in probes, including that they sit inside the middleware chain so app-wide auth applies to them.

## Shorter — 2374 → 1774 lines

Cut **only where a docs page already carries the same material**, so nothing is lost, only relocated:

| package | lines | what went |
| --- | --- | --- |
| mcp | 560 → **176** | auth patterns + ASCII flow diagram, boot-sequence and dispatch diagrams, transport diagram, troubleshooting table, and an alternative the README itself called not-recommended — all in the 1031-line guide |
| schema | 283 → **132** | per-adapter internals, two resolution orders, a full Joi adapter implementation, a wiring diagram restating the one-line pitch |
| grpc | 206 → **121** | error mapping, streaming, boot failures, introspection. Kept the protocol-support table — that's the decision |
| devtools-kit | 113 → **103** | "Why a separate kit package" restated the opening; the peer-optional bullets explain npm, not this package |

Everything else was already the right length. No package lacks a README.

## Two audit claims I did not act on

The audit said the `middleware` singular alias still exists and that devtools-kit is at 0.1.2. Neither holds: the alias is gone on `main` (#595), and devtools-kit is 7.0.1 — so those ranges were stale, not install-breaking as reported. Fixed to 7.x anyway.

## Self-inflicted, caught before commit

Replacing schema's adapter list, I wrote a table listing a `@forinda/kickjs-schema/standard` subpath that does not exist — and which duplicated the accurate subpath table directly below it. Checking against the package's `exports` map killed both problems.

CLI suite green: 74 files, 770 tests.

## Summary by CodeRabbit

- **New Features**
  - Documented built-in `/health/live` and `/health/ready` endpoints, including readiness checks and opt-out configuration.
  - Added documentation for the `fullstack` project template and expanded CLI generator commands.
  - Updated testing guidance for runtime-specific app handling and dependency overrides.

- **Documentation**
  - Refreshed CLI package examples and supported package lists.
  - Simplified MCP, gRPC, schema, and development-kit guides with links to external detailed references.
  - Updated README examples and package recommendations.


forinda:
Both valid, both fixed in 5171a865 — verified against source rather than taken at face value.

**grpc `see above`** — correct, and self-inflicted: the Options block's `contributors: []` comment referenced the Context Contributors section my own cut removed. Points at the guide now.

**mcp `@McpTool` gating** — correct, and the more useful of the two. `mcp.adapter.ts:159` gates on `mode === 'explicit' && !meta`, so in `auto` mode an undecorated route *is* exposed when it matches `include`/`exclude`; the adapter's own comment says "the decorator may be absent entirely". Stating the firewall claim unconditionally invited exactly the wrong conclusion in the mode where it doesn't hold.

The paragraph now scopes the gate to `explicit`, says plainly that `auto` removes it, and names `@McpTool({ hidden: true })` as the opt-out — which `:160` checks *before* the mode branch, so it works in either mode.

forinda:
Valid, and worse than reported — fixed in the latest commit.

The sentence was mine, written while cutting this README, and it was wrong twice over:

- **`parse` is not in the interface.** `KickSchema` (`src/types.ts:16`) is `safeParse` + `toJsonSchema`. `isKickSchema` (`src/detect.ts:16-17`) checks for exactly those two at runtime.
- **An adapter is not a bare function.** `registerAdapter` takes a `SchemaAdapter` of `{ name, detect, wrap }` (`src/types.ts:21-25`).

It also sat directly beneath the README's own — correct — `KickSchema` interface block, so the page contradicted itself on a single screen.

Now states the contract plainly (both members, and that `toJsonSchema` is what feeds OpenAPI) and carries a worked Joi adapter using the same imports and helper names as `docs/guide/schema.md:178`, so moving between the two doesn't read as two different APIs.

Issue #590, opened 2026-08-31
Request bodies parse differently per runtime for content types other than application/json


forinda:
The same request body is parsed three different ways depending on the configured runtime. A request that works on one engine is a `415`, an empty body, or a raw string on another — with nothing in the types or the config to signal it.

Found while adding the runtime matrix in #586; out of scope there, so filing separately.

## Measured

`POST /api/v1/echo`, handler returns `ctx.body`, app built with `createTestApp` (default middleware — `express.json()` under Express, native parsing elsewhere).

| body sent | Express | Fastify | h3 |
| --- | --- | --- | --- |
| `application/x-www-form-urlencoded` (`a=1&b=2`) | `200` — body **undefined** | **`415`** | `200` — `{ a: '1', b: '2' }` |
| `application/merge-patch+json` (valid) | `200` — body **undefined** | **`415`** | `200` — `{ a: 1 }` |
| `application/merge-patch+json` (malformed) | `200` — body undefined | `415` | `200` — body is the **raw string** `'{"a":'` |
| `text/plain` | `200` — body undefined | `200` — `'hello'` | `200` — `'hello'` |
| no `Content-Type` | `200` — body **undefined** | **`415`** | `200` — `{ a: '1' }` |

`application/json`, with or without parameters, is consistent on all three — that part is covered by #586.

## Why each behaves that way

- **Express** parses only what its middleware list handles. The default is `express.json()`, so anything else arrives unparsed and `ctx.body` is `undefined`. Adding `express.urlencoded()` changes it — this one is at least *configurable*.
- **Fastify** refuses a content type it has no parser registered for, with `415`. That is defensible on its own, but it is the only engine that rejects rather than ignores.
- **h3** routes everything unrecognised through `destr`, so it parses form bodies and `+json` types opportunistically — and for malformed input in that branch hands the handler a **raw string** where a parsed object is expected.

## Why it matters

The runtime is meant to be swappable (`bootstrap({ runtime })`). These make it not swappable for any app that accepts a body outside `application/json`:

- Express → Fastify: every form post starts returning `415`.
- Express → h3: bodies begin parsing that previously did not, so handlers that guarded on `!ctx.body` change behaviour silently.
- h3 hands a `string` to a handler whose type says object, for a malformed `+json` body. That is the same class as the bug fixed in #586, in the branch that fix deliberately did not widen into.

## What a fix would need to decide

Not obvious, which is why this is an issue rather than a PR:

[...]

PR #606, opened 2026-08-31
fix(kickjs): stop the web entry swallowing a malformed request body

### LINKED ISSUE #605: Web entry answers 200 for a malformed JSON body, handler runs on undefined
The `./web` entry (`h3WebRuntime`, used for edge / Bun / Deno) answers **200** for a malformed JSON body and runs the handler with `ctx.body === undefined`.

This is the same defect fixed for the node h3 runtime in #586. That fix did not reach the web entry, which has its own body-reading path.

## Measured

`@forinda/kickjs` 7.4.0, current `main`. `POST` with `content-type: application/json` and body `{"a":`  :

```
status=200  handlerRan=true  ctx.body=undefined
```

Express and Fastify both answer **400** for the same request, and so does the node h3 runtime since #586.

## Cause

`packages/kickjs/src/http/web/handler.ts:120-129`:

```ts
} else if (BODY_METHODS.has(request.method)) {
  const contentType = req.headers['content-type'] ?? ''
  if (contentType.includes('application/json')) {
    req.body = await request.json().catch(() => undefined)   // ← line 123
  } else if (contentType.includes('application/x-www-form-urlencoded')) {
    const text = await request.text().catch(() => '')
    req.body = Object.fromEntries(new URLSearchParams(text))
  } else if (contentType.startsWith('text/')) {
    req.body = await request.text().catch(() => undefined)
  }
}
```

`.catch(() => undefined)` cannot tell an **absent** body from an **unparseable** one. That is the exact reasoning in #586, whose fix decides on `content-length` / `transfer-encoding` instead of on whether parsing threw (`runtimes/h3.ts:182-199`):

```ts
const sentBody = encoding !== undefined || (length !== undefined && length !== '0')
if (sentBody) {
  try {
    req.body = await readBody(event, strict ? { strict: true } : {})
  } catch (err) {
    throw HttpException.badRequest(...)
  }
}
```

## Why it matters

Worse than the wrong status: the client is told it succeeded, and the handler executes against data that never parsed. A create endpoint writes whatever its defaults are. This is the failure mode #586 exists to remove — it is simply still present on the entry used for edge, Bun and Deno deployments, which is where it is least likely to be noticed.

The form-urlencoded branch has a milder version of the same problem: `.catch(() => '')` turns a failed read into an empty object rather than an error.

## Repro

```ts
import * as h3v2 from 'h3'
import { h3WebRuntime } from '@forinda/kickjs/h3-web'

// mount any POST route whose handler echoes ctx.body, then:
const res = await app.fetch(
  new Request('http://test/api/things', {
    method: 'POST',
    headers: { 'content-type': 'application/json' },
    body: '{"a":',
  }),
)
// actual:   200, handler ran, ctx.body === undefined
// expected: 400, handler did not run
```

## Suggested fix

Apply #586's shape to `web/handler.ts`: decide whether a body was sent from the request rather than from whether parsing threw, and let a parse failure raise `HttpException.badRequest` instead of being swallowed. `Request.body !== null` is the web-standard equivalent of the `content-length` / `transfer-encoding` check.

Worth pinning in the runtime matrix at the same time — `packages/testing/__tests__/body-parse-errors.test.ts` covers express / fastify / h3 but not the web entry, which is why this survived #586.

Found while researching #590. Filed separately because it is a defect with a known fix, where #590 is an open design question.

forinda:
Closes #605.

## The bug

`h3WebRuntime` — the `./web` entry for edge, Bun and Deno — read JSON with:

```ts
req.body = await request.json().catch(() => undefined)
```

That catch cannot tell an **absent** body from an **unparseable** one, so broken JSON produced a **200** with the handler running against `undefined`.

Same defect #586 removed from the node h3 runtime. It survived there because the web entry has its own body-reading path, and `body-parse-errors.test.ts` covers express / fastify / h3 but not this entry.

Worse than the wrong status: the client is told it succeeded and the handler executes on data that never parsed — a create endpoint writes whatever its defaults are. On the entry used for edge deployments, where it's least likely to be noticed.

## The fix

Read the body once as text, and decide from the request rather than from whether parsing threw:

- **no body, or an empty one** → absent. `ctx.body` stays undefined, request succeeds. This is the case the original catch existed for, and the same call the node runtime makes on `content-length: 0`.
- **present but unparseable** → `HttpException.badRequest`, so all four runtimes answer 400.

## One thing worth reviewing

The parse failure is **held, not thrown at the point of discovery**. The response driver is constructed *after* the body read, so throwing there escapes the route entirely and surfaces as a 500. It's rethrown as the pipeline's first act, where the existing error path maps it.

The test caught this: it went `201` (bug) → `500` (thrown too early) → `400` (correct) across the two attempts. Without a test written first I'd likely have shipped the 500.

## Behaviour changes riding along

Both toward parity with the node runtime:

- a failed `x-www-form-urlencoded` read leaves `ctx.body` undefined rather than `{}`
- an empty body of any type reads as absent rather than an empty object

## Verification

Two tests on the web runtime — malformed JSON rejected with the handler not running, and a POST with no body still succeeding. Both written before the fix; the first failed with `expected 201 to be 400`. Sabotage-checked by reverting only the rethrow, which fails it again.

Suite green: 20 build tasks, 38 test tasks.

Does not touch #590 — that remains the open design question about which content types the framework claims and whether an unsupported one is ignored or rejected.

## Summary by CodeRabbit

- **New Features**
  - Unified request-body handling across supported runtimes.
  - JSON, `+json`, URL-encoded forms, and text bodies are parsed consistently.
  - Unsupported body types now return a clear 415 response with accepted formats.
  - Empty requests remain supported regardless of declared content type.

- **Bug Fixes**
  - Malformed JSON now returns 400 without invoking the route handler.
  - Empty and absent bodies are correctly distinguished from invalid payloads.

- **Documentation**
  - Added migration guidance for updated body-parsing behavior.


forinda:
Fixed in 36a8b522. Both source findings were real; one was a behaviour bug, not just types.

**Typecheck (fastify.ts).** Correct, and my miss: I ran `pnpm build` and `pnpm test` on this work but not `typecheck`. tsdown doesn't type-check — the CI gate is `tsc --noEmit`. `FastifyAppLike` now declares `addContentTypeParser` and `FastifyRequestLike` declares `headers`, with a named `FastifyBodyParser` shape so the callbacks are checked at the call site rather than cast away. These parsers decide what `ctx.body` is on that runtime, so they're worth checking.

Also widened `normalizeMediaType` / `classifyMediaType` / `unsupportedMediaTypeError` to take `string | string[] | undefined` — Node's `IncomingHttpHeaders` and Fastify's normalised map both allow arrays, and taking the union at the boundary beats casting at every call site.

**Empty chunked body (h3.ts) — the good catch.** This was a genuine inconsistency with the same commit's own rule: 415 is for a body that cannot be *read*, never for the absence of one. Fastify and the web entry already checked emptiness; h3 and the Express guard decided from framing headers, so an empty chunked post was rejected.

h3 now rejects only after `readRawBody` shows content. The Express guard waits for a byte to arrive on a chunked request, since Express has no parsed body to inspect for an unsupported type.

Regression test added across all three runtimes, as suggested — it failed on **h3 and Express** before the change, so the finding was broader than the one file it was filed against.

**Not actioned:** the suggestion to add `src/env.ts` + `loadEnv()` to `packages/testing`. That guideline is for scaffolded applications; the testing package is a library harness with no env schema of its own.

MD040 fence tagged `text`.

All four CI gates run locally this time: build, typecheck, test, format.

forinda:
Not actioned — the finding applies an application rule to a library package. Evidence rather than assertion:

- **No package under `packages/` has an `src/env.ts`.** Not one of the 18.
- The only two `import './env'` occurrences in repo source are the framework telling *adopters* to add it: an error-message template in `packages/kickjs/src/core/kick-errors.ts:157`, and the `kick doctor` check in `packages/cli/src/commands/doctor.ts:593`, which keys on the adopter's `ctx.cwd` and their `indexPath`.
- **`packages/testing` has no env surface at all** — no `loadEnv`, no `envSchema`, no `defineEnv`. There is no schema for a `src/env.ts` to load.

The guideline's own wording is application-scoped: it says the import must come *"before `bootstrap()` runs"* and that *"the CLI generators wire both halves automatically"*. `createTestApp` constructs an `Application` for a test; it is not an app with an entry point and a config schema.

That said, the guideline didn't say so explicitly, which is why it read as repo-wide. I've scoped it in `CLAUDE.md` so the same false positive doesn't recur for a human or a bot:

> **Env wiring** (applies to **generated applications**, not to the workspace's own library packages under `packages/` — none of which have or need an `src/env.ts`): …

PR #607, opened 2026-09-01
feat(kickjs)!: one body-parsing policy for every runtime

forinda:
Stacked on #606 (the web-entry fix) — merge that first.

Implements steps 2 and 3 of the plan in #590. Step 4 is deliberately **not** here; see the end.

## Before / after

| body sent | v7 Express | v7 Fastify | v7 h3 | now, all three |
| --- | --- | --- | --- | --- |
| `application/x-www-form-urlencoded` | `undefined` | **415** | parsed | **parsed** |
| `application/merge-patch+json` | `undefined` | **415** | parsed | **parsed as JSON** |
| malformed `+json` | `undefined` | 415 | **raw string** | **400** |
| `text/plain` | `undefined` | `'hello'` | `'hello'` | **`'hello'`** |

`bootstrap({ runtime })` is meant to be swappable. It was not for any app accepting a body outside `application/json`: Express → Fastify turned every form post into a 415, and Express → h3 started parsing bodies that previously did not, silently changing handlers that guarded on `!ctx.body`.

## The policy

`http/body-policy.ts` decides once; all four runtimes follow it, web entry included.

- `application/json`, `application/*+json` → strict JSON, malformed is 400
- `application/x-www-form-urlencoded` → object
- `text/*` → raw string
- `multipart/*` → unchanged, the upload path consumes the stream

**`+json` is JSON by specification, not by liberty.** RFC 6838 §4.2.8: a type "MUST NOT be given names incorporating suffixes for structured syntaxes they do not actually employ." RFC 6839 §2 exists so receivers can do "generic processing of the underlying representation." Spring, ASP.NET Core and Hono match the suffix by default. It also means the framework can finally read back the `application/problem+json` it emits for every problem response.

**`text/*` is never JSON-parsed, deliberately.** `text/plain` is one of three CORS-safelisted content types — it crosses origins with no preflight. JSON-parsing it re-opens the simple-request CSRF that requiring `application/json` closes. h3's own source carries this warning.

## Per engine

- **Express** — default chain gains `urlencoded` and `text`; the JSON parser gets **both** `application/json` and `application/*+json`, because `type-is` will not match plain JSON against the wildcard alone (a length guard rejects it), so passing only the wildcard silently stops parsing ordinary JSON.
- **Fastify** — content-type parsers for the same set. Deliberately **not** a `'*'` catch-all: that would swallow multipart, and `@fastify/multipart` *is* the multipart parser, registered conditionally. Its parse errors are raised as `HttpException.badRequest` — a bare `SyntaxError` has no status and mapped to 500, which the matrix caught.
- **h3** — reads raw and applies the policy rather than calling `readBody`, whose own dispatch was the source of the divergence (`===` on `application/json`, `destr` for everything else).

## One finding worth flagging

`createTestApp` named its own middleware list, which put the Application on its **user-declared** branch — so a test app parsed only `application/json` while the same app in production parsed the full set. **The harness was quietly exercising a different pipeline from the one deployed.** That is why the Express rows failed on first run. It now leaves `middlewares` undefined and lets the Application apply its own defaults; the native-body guard it was duplicating already lives there.

## Still open — step 4

A content type **outside** the supported set (`application/xml`, say) is ignored on Express and h3 but still `415`'d by Fastify. Ignore-vs-reject is a breaking change either way, so it needs your call rather than my default. My recommendation in #590 is unchanged — 415 when a body was actually sent, treating empty as absent, which is what Spring and ASP.NET both do — but it breaks Express apps that rely on silent ignoring, so it should be a deliberate decision, not a side effect of this PR.

Documented in the v8 migration guide, including that open divergence.

## Verification

New matrix over all three runtimes: JSON, `+json`, malformed `+json`, form-urlencoded, `text/plain` as a string, malformed JSON, POST with no body. Five of the 21 cases failed before the fix.

Suite green: 20 build tasks, 38 test tasks.

### DIFF
--- a/.changeset/great-swans-agree.md
+++ b/.changeset/great-swans-agree.md
@@ -1,0 +1,63 @@
+---
+'@forinda/kickjs': minor
+'@forinda/kickjs-testing': patch
+---
+
+Request bodies parse the same way on every runtime.
+
+Each engine brought its own library's opinion about content types, so the same
+request produced three different results (#590):
+
+| body sent                           | Express     | Fastify   | h3             |
+| ----------------------------------- | ----------- | --------- | -------------- |
+| `application/x-www-form-urlencoded` | `undefined` | **415**   | parsed         |
+| `application/merge-patch+json`      | `undefined` | **415**   | parsed         |
+| malformed `+json`                   | `undefined` | 415       | **raw string** |
+| `text/plain`                        | `undefined` | `'hello'` | `'hello'`      |
+
+`bootstrap({ runtime })` is meant to be swappable. It was not, for any app
+accepting a body outside `application/json`: Express → Fastify turned every
+form post into a 415, and Express → h3 started parsing bodies that previously
+did not, silently changing handlers that guarded on `!ctx.body`.
+
+One policy now decides, in `http/body-policy.ts`, and all four runtimes follow
+it — including the web entry for edge, Bun and Deno:
+
+- `application/json` and `application/*+json` — strict JSON; malformed is 400
+- `application/x-www-form-urlencoded` — parsed to an object
+- `text/*` — the raw string
+- `multipart/*` — unchanged, the upload path consumes the stream itself
+
+**`+json` is JSON by specification, not by liberty.** RFC 6838 §4.2.8: a media
+type "MUST NOT be given names incorporating suffixes for structured syntaxes
+they do not actually employ"; RFC 6839 §2 exists so receivers can do "generic
+processing of the underlying representation". Spring, ASP.NET Core and Hono all
+match the suffix by default. It also means the framework can read back the
+`application/problem+json` it emits for every problem response.
+
+**`text/*` is never JSON-parsed, deliberately.** `text/plain` is one of three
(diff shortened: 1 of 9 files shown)


forinda:
Updated for your call: unsupported types are now **rejected with 415**, not ignored.

A body the framework cannot read answers 415 with an `Accept` header naming what it accepts, on every runtime. The rejection is for a body that cannot be *read*, never for the absence of one — a bodyless POST succeeds whatever its declared type, matching what Spring and ASP.NET both do.

Fastify needed that spelled out: it invokes a content-type parser even for an empty payload, so without the guard a bodyless POST carrying an unrelated type was rejected. The matrix caught it.

Express needed a guard after its parsers, since it has no concept of an unsupported type — a parser whose `type` misses just calls `next()` and leaves `req.body` undefined, indistinguishable from nothing being sent.

This makes the PR breaking for Express apps that accepted an unusual content type and ignored the body; changeset bumped to major and the migration guide says so. Closes #590 rather than deferring step 4.

Issue #609, opened 2026-09-01
Generated module's eager glob misses decorated classes outside three filename suffixes


forinda:
The eager-load glob in a generated module only covers three filename suffixes, so a decorated class in any other file is never imported, never registers, and fails at resolve time with `No provider for`. The fix an adopter has to discover is an explicit side-effect import — which is exactly what the glob exists to avoid.

## What is generated

`packages/cli/src/generators/templates/module-index.ts:189-192`:

```ts
import.meta.glob(
  ['./**/*.controller.ts', './**/*.service.ts', './**/*.repository.ts', '!./**/*.test.ts'],
  { eager: true },
)
```

with the comment: *"Eagerly load decorated classes so @Controller()/@Service()/@Repository() decorators register in the DI container."*

## The gap

Only those three suffixes are matched. A decorated class in a file named anything else is silently skipped:

- `*.usecase.ts`
- `*.policy.ts`
- `*.mapper.ts`
- `*.handler.ts`
- `*.factory.ts`
- `*.gateway.ts`

A class that is **routed** gets pulled in transitively by the controller that imports it, so the gap only shows up for a class nothing else imports — an unrouted `@Service` reached solely through `container.resolve()` or `@Autowired`. That is the case most likely to be written by hand and least likely to be caught by the generated tests.

## Why it is worth fixing rather than documenting

The glob's whole purpose is that an author should not have to remember to import a file for its decorator side effect. A glob that covers *some* files reproduces the original problem in a smaller and less obvious form: it works for the three names the generator happens to emit, and fails for the fourth name an adopter picks. Nothing warns.

## Options

1. **Widen to the module folder** — `./**/*.ts` with `!./**/*.test.ts` (and probably `!./**/*.dto.ts`). Simple, no list to keep in step, at the cost of importing files that carry no decorators.
2. **Keep a suffix list, and make it visible.** Emit the list into the scaffolded comment so an adopter adding `*.usecase.ts` can see they must extend it. Cheaper, still a trap.
3. **Detect and warn.** Have `kick doctor` flag a decorated class under a module folder that no glob pattern in that module's index matches.

Option 1 seems right — the exclusion list is short and the cost is a few extra module evaluations at boot, against a failure mode that surfaces as a runtime `No provider for` with no pointer to the cause.

Whichever is chosen, `docs/guide/modules.md` describes this glob and should say plainly which files it does and does not reach.

Found alongside #608 while testing DI resolution from a test container.

PR #610, opened 2026-09-01
fix: seed isolated containers, and widen the generated eager glob

### LINKED ISSUE #608: Container.create() returns a container with no decorator registrations, breaking isolated: true
`Container.create()` returns a container that no decorator registration ever reaches, so every `@Service` / `@Repository` / `@Controller` class is missing from it. `createTestApp({ isolated: true })` and `createTestPlugin` both use it, and the generated project docs recommend it.

## Repro

```ts
@Service()
class PricingService {
  quote() { return 42 }
}

const { container } = await createTestApp({ modules: [PingModule()], isolated: true })
container.resolve(PricingService)
```

```
KICK001: No provider for PricingService
```

The same resolve succeeds on the default (non-isolated) container.

## Cause

`@Service()` reaches `registerInContainer` (`packages/kickjs/src/core/decorators.ts:104`), which registers into `containerRef` — the global singleton — at **class-definition time**, and keeps a copy in the module-level `allRegistrations` map.

That map exists precisely so a fresh container can be re-seeded. `Container.reset()` uses it (`decorators.ts:46-53`):

```ts
Container._onReset = (container: any) => {
  containerRef = container
  for (const [, { target, scope }] of allRegistrations) {
    if (!container.has(target)) container.register(target, target, scope)
  }
}
```

`Container.create()` has no equivalent — it is literally `new Container()` (`packages/kickjs/src/core/container.ts:261-263`), with no replay and no hook. So an isolated container starts empty of everything the decorators registered.

## Why it matters

`isolated: true` is documented as "safe for concurrent tests" (`packages/testing/src/index.ts:96-99`). It is — it just silently empties the DI graph while doing it, and the failure surfaces later as `No provider for X` in a test that looks like it should pass.

It is also actively taught. `packages/cli/src/generators/templates/project-docs.ts` recommends `Container.create()` for test isolation in five places, including:

- `:161` — "**Repos under tests**: `Container.create()` for isolation — never `new Container()` or `getInstance().reset()`."
- `:321` — "**Test isolation** — default to `Container.create()` for fresh DI state."

So every scaffolded project ships advice that produces this failure the moment a test resolves a decorator-registered class.

`createTestPlugin` takes the same path (`packages/testing/src/index.ts:489`), and defaults `isolated` to **true**.

## Suggested fix

Give `create()` the same replay `reset()` has, without letting it hijack the global `containerRef` — an isolated container should be seeded, not adopted:

```ts
// decorators.ts
Container._onCreate = (container: any) => {
  for (const [, { target, scope }] of allRegistrations) {
    if (!container.has(target)) container.register(target, target, scope)
  }
}
```

and call it from `Container.create()`. That fixes `createTestApp({ isolated: true })`, `createTestPlugin`, and every adopter following the generated docs, in one place.

Worth pinning in `packages/testing/__tests__` both ways: a decorator-registered service resolves from an isolated container, and an isolated container does not leak into or out of the global one.

forinda:
Closes #608 and #609 — both found while checking your report about resolving a service from a test container.

## #608 — `isolated: true` resolved nothing

```
KICK001: No provider for PricingService
```

Decorators register at **class-definition time** into whichever container was global then, and keep a copy in `allRegistrations` so a fresh container can be re-seeded. `Container.reset()` replays that map (`decorators.ts:46`). `Container.create()` was literally `new Container()` — no replay, no hook. So an isolated container was missing every `@Service`, `@Repository` and `@Controller` in the process, while being documented as "safe for concurrent tests".

**It also ships as advice.** `project-docs.ts` recommends `Container.create()` for test isolation in five places, and `createTestPlugin` defaults `isolated` to true. Every scaffolded project carried guidance that produces this failure.

`create()` now seeds through a **separate `_onCreate` hook that deliberately does not reassign `containerRef`**. That distinction is the whole fix: `_onReset` both seeds *and* adopts, because after a reset the new container really is the global one. An isolated container must only get the first half — otherwise it would capture every registration made after it was created, turning a missing-provider bug into a cross-test-contamination bug. The third test pins exactly that.

## #609 — the eager glob missed most filenames

The generated module globbed `*.controller.ts`, `*.service.ts`, `*.repository.ts`. A decorated class in `*.usecase.ts`, `*.policy.ts` or `*.mapper.ts` was never imported, so its decorator never ran.

A *routed* class gets pulled in transitively by the controller that imports it, which is why this only bit an **unrouted** service reached through `resolve()` or `@Autowired` — your exact symptom, and the case least likely to be covered by the generated tests.

Now `./**/*.ts` minus tests and `.d.ts`. I went with the broad glob rather than a longer suffix list because a list only ever covers the filenames the generator itself emits: it works until an adopter picks a fourth name, and nothing warns. That is the original problem in a smaller, less obvious form. `docs/guide/modules.md` documented the old glob and is updated with it.

## Verification

Three tests in `packages/testing/__tests__/isolated-container.test.ts`: the isolated container resolves a decorator-registered service, agrees with the shared container, and leaks in neither direction. Two of the three fail with the `_onCreate` call removed.

Gates: 20 build, 24 typecheck, 38 test, docs build clean.

## Still open from your report

I could not reproduce the third item — that the test-app options require setting the healthcheck explicitly. `/health/live` and `/health/ready` both serve with a bare `createTestApp({ modules })`, and `health` is optional in the type (`packages/testing/src/index.ts:46`, picked from `ApplicationOptions.health?: boolean`). If you have the exact call and error I will chase it; I may have been testing the wrong surface.

## Summary by CodeRabbit

- **Bug Fixes**
  - Isolated containers now correctly resolve classes registered with service, repository, and controller decorators.
  - Generated modules now discover decorated TypeScript files recursively, including use cases, policies, and mappers, while excluding tests and declaration files.
  - Isolated containers remain separate from the shared application container, preventing registration leaks.

- **Documentation**
  - Updated module-loading guidance and examples to reflect recursive TypeScript file discovery.

### DIFF
--- a/.changeset/olive-frogs-attend.md
+++ b/.changeset/olive-frogs-attend.md
@@ -12,7 +12,7 @@ be re-seeded. `Container.reset()` replays that map; `create()` was literally
 `new Container()`, so an isolated container was missing every `@Service`,
 `@Repository` and `@Controller` in the process:
 
-```
+```text
 KICK001: No provider for PricingService
 ```
 
--- a/.changeset/olive-frogs-attend.md
+++ b/.changeset/olive-frogs-attend.md
@@ -1,0 +1,45 @@
+---
+'@forinda/kickjs': patch
+'@forinda/kickjs-cli': patch
+---
+
+Fix isolated containers resolving nothing, and a glob that missed half a module.
+
+**`Container.create()` returned a container with no decorator registrations
+(#608).** Decorators register at class-definition time into whichever container
+was global then, keeping a copy in `allRegistrations` so a fresh container can
+be re-seeded. `Container.reset()` replays that map; `create()` was literally
+`new Container()`, so an isolated container was missing every `@Service`,
+`@Repository` and `@Controller` in the process:
+
+```
+KICK001: No provider for PricingService
+```
+
+`createTestApp({ isolated: true })` and `createTestPlugin` both take that path,
+and `createTestPlugin` defaults `isolated` to true. Worse, the scaffolded
+project docs recommend `Container.create()` for test isolation in five places —
+so the advice every new project ships produced this failure the moment a test
+resolved a decorator-registered class.
+
+`create()` now seeds the same way `reset()` does, through a separate
+`_onCreate` hook that deliberately does **not** reassign the decorators'
+`containerRef`. Seeding and adopting are different things: an isolated
+container should receive what has been registered so far, not become the
+destination for everything registered next. Pinned by a test that the isolated
+container resolves the service, agrees with the shared one, and still does not
+leak either way.
+
+**The generated module's eager glob only matched three suffixes (#609).** It
+covered `*.controller.ts`, `*.service.ts` and `*.repository.ts`, so a decorated
+class in `*.usecase.ts`, `*.policy.ts` or `*.mapper.ts` was never imported,
+never registered, and failed later as `No provider for X`. A routed class was
+pulled in transitively by its controller, so the gap only showed for an unrouted
+one reached through `resolve()` or `@Autowired` — the case least likely to be
+covered by the generated tests.
+
+It is now `./**/*.ts` minus tests and declarations. A suffix list only ever
+covers the filenames the generator itself emits, which reproduces the original
+problem in a smaller and less obvious form: it works until an adopter picks a
+fourth name, and nothing warns. `docs/guide/modules.md` describes the same glob
+and is updated with it.
--- a/.changeset/olive-frogs-attend.md
+++ b/.changeset/olive-frogs-attend.md
@@ -30,6 +30,28 @@ destination for everything registered next. Pinned by a test that the isolated
 container resolves the service, agrees with the shared one, and still does not
 leak either way.
 
+**Isolated containers also leaked, in the other direction.** `registerInstance`,
+`registerFactory` and `resolve`'s singleton cache all wrote to the persistent
+store — the `globalThis` map that exists so the GLOBAL container survives HMR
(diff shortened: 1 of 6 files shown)


forinda:
Real, and worse than reported — fixed in d33c73de.

Reproduced first:

```ts
Container.create().registerInstance('probe', { v: 1 })
Container.reset()
Container.getInstance().has('probe')   // true
```

**Three writers, not one.** `registerInstance` was the one you found, but `registerFactory` and `resolve()`'s singleton cache write to the same persistent store. My first patch guarded two of them and the factory test still failed — which is how the third surfaced.

**And it leaks in both directions.** `registerFactory` *reads* the store to reuse an HMR-cached instance, so a freshly created isolated container could adopt global state and start out polluted. Same contamination, opposite direction. Both sides are now gated.

Persistence is scoped by an explicit `isGlobal` flag set in `getInstance()` and `reset()`, rather than inferred from `Container.instance === this` — the inference is correct today but silently wrong the moment anything else constructs a container.

**On timing:** this predates the PR, but the PR is what makes it reachable. Before it, an isolated container resolved nothing, so nobody got far enough to leak from one. Fixing the missing-provider bug without this would have traded it for cross-test contamination — quieter and harder to diagnose than the error it replaced.

Three regression tests, as you suggested, including your exact scenario: an override registered through `createTestApp({ isolated: true, overrides })` is absent from the global container after a reset; a resolved factory instance is too; and the isolated container is still a distinct object. The first two fail without the change.

Gates: 20 build, 24 typecheck, 38 test.

forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

PR #617, opened 2026-09-01
chore: ship the problem-details change as a minor, not a major

### LINKED ISSUE #123: feat(otel): batteries-included preset (NodeSDK + OTLP + pino correlation + Sentry + shutdown flush)
Split from #121. The \`sensitiveKeys\` half landed; this tracks the full observability preset.

## Scope

Ship a second entrypoint (e.g. \`@forinda/kickjs-otel/preset\` or \`registerFullStack({...})\` helper) that wires:

- \`NodeSDK\` with configurable exporters (OTLP / console / none)
- Pino trace-correlation mixin (pulls active span's traceId / spanId into every log record)
- Sentry integration (opt-in, tied to OTel span context)
- Service resource attributes (\`service.name\`, \`service.version\`, \`deployment.environment\`)
- Graceful shutdown flush in the SIGTERM handler

The current HTTP-only export stays for users who only want route spans.

## Why separate

Adds multiple peer deps (\`@opentelemetry/sdk-node\`, \`@opentelemetry/exporter-trace-otlp-http\`, \`@opentelemetry/instrumentation-pino\`, \`@sentry/node\`) and needs real-world testing against Jaeger / Tempo / Datadog / Honeycomb to avoid shipping a preset that \"mostly works.\" Better as a dedicated PR than a bundle.

Filed from #121.

forinda:
8.0.0 released a day ago, so the major changeset put this at 9.0.0 — which would say far more about the release cadence than about the change.

The response shape does change, so semver argues for a major, and the changeset now says that outright rather than leaving the choice unexplained. The window in which an adopter has pinned `{ message }` against 8.0.0 specifically is about one day wide.

The migration guide framed it under v8's breaking changes; it is retitled to "In 8.1" and the at-a-glance row is tagged, so a reader upgrading from v7 does not go looking for it in the 8.0 notes.

Claude-Session: https://claude.ai/code/session_011wf3GNMSJUfME42a93d5j6

## Type

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] CI / Tooling

## Changes

-

## Checklist

- [ ] `pnpm build` passes
- [ ] `pnpm test` passes
- [ ] `pnpm format:check` passes
- [ ] Docs updated (if applicable)
- [ ] New exports added to `index.ts` and `package.json` exports map (if applicable)

## Summary by CodeRabbit

- **New Features**
  - Framework errors now use the standardized `application/problem+json` format, including type, title, status, and detail fields.
  - Validation errors now follow the same standardized response format.

- **Documentation**
  - Updated the v7-to-v8 migration guide to clarify that standardized error responses were introduced in version 8.1.

### DIFF
--- a/.changeset/proud-lions-repeat.md
+++ b/.changeset/proud-lions-repeat.md
@@ -1,5 +1,5 @@
 ---
-'@forinda/kickjs': major
+'@forinda/kickjs': minor
 ---
 
 Every error the framework serialises is RFC 9457 problem details.
@@ -28,6 +28,12 @@ bodies move with it.
 
 **Migrating:** assert on `body.detail` rather than `body.message`. Nothing is
 lost, it is renamed to the RFC's field.
+
+Released as a **minor** rather than a major, deliberately. The response shape
+does change, so semver would argue for a major — but 8.0.0 shipped a day
+earlier, and moving to 9.0.0 for this would say something much louder about the
+release cadence than about the change. The window where an adopter has pinned
+`{ message }` against 8.0.0 specifically is roughly that one day.
 
 The error handler's own contract previously said `HttpException` "keeps the
 existing `{ message, errors? }` shape for backward compatibility". That clause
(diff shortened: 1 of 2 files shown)


forinda:
Duplicate — this change already landed on main via #615. Closing.

PR #635, opened 2026-09-03
feat: route flags reach the OpenAPI spec and the DevTools dashboard

forinda:
Phase 4 of [`route-flags-design.md`](https://github.com/forinda/kick-js/blob/main/route-flags-design.md)
— the readers. Independent of #634 (phase 2); both build on the phase 1 core in #631.

## `getRouteFlags(controllerClass, handlerName)`

Resolves a route's flags **outside a request** — the same method-over-class result
`ctx.route.flags` carries, for consumers that see a controller and a method name rather than a
live ctx: an adapter's `onRouteMount`, spec generation, tooling.

## Swagger: `publicFlag`

```ts
export const Public = defineRouteFlag('auth.public')

SwaggerAdapter({ bearerAuth: true, publicFlag: 'auth.public' })
```

```ts
@Public                      // whole controller documented public
@Controller()
class WebhooksController {
  @Get('/health')  health(ctx) {}

  @Public(false)             // documented as secured again
  @Post('/admin')  admin(ctx) {}
}
```

**The flag name is the option, not a constant** — that's the point. The framework deliberately
names no flags, so one project's `auth.public` is another's `public` or `security.none`. Naming it
lets the spec read the same declaration the runtime reads, instead of asking for a second
annotation that drifts from it. A list accepts several:

```ts
SwaggerAdapter({ bearerAuth: true, publicFlag: ['auth.public', 'health.probe'] })
```

Resolution order: `@ApiPublic` → `securityResolver` → **`publicFlag`** → `@ApiSecurity` /
`@ApiBearerAuth` (method, then class). So an explicit resolver still wins, and a flag still
overrides class-level security.

Swagger needed no new plumbing beyond the option — `securityResolver` already existed for exactly
this kind of bridge, and `publicFlag` is its common case spelled as config. For anything richer
(reading a flag's *value*, combining two), `securityResolver` + `getRouteFlags` still covers it,
and there's a test for that too.

## DevTools: flags in the route list

`GET /_debug/routes` now carries `flags` per entry, and the dashboard's Routes tab shows them in a
Flags column:

```json
{
  "method": "GET",
  "path": "/api/v1/health",
  "controller": "HealthController",
  "handler": "live",
  "middleware": [],
  "flags": { "auth.public": true }
}
```

Only flags **in force** appear — one turned off at the method is absent, not `false` — so the
dashboard agrees with every other consumer. Valued flags render as `rate.limit={"rpm":10}`.

That makes "why does this endpoint not require auth" answerable from the route list instead of by
reading the controller.

## Verification

6 new tests across swagger and devtools: the named lookup, a list of names, class inheritance with
a method-level opt-out, an explicit resolver overriding the flag, the `getRouteFlags` +
`securityResolver` form, and `/_debug/routes` reporting resolved flags.

Repo suite: **311 files, 3248 tests, all passing.**

## Note

`docs/guide/route-flags.md` arrives with #634, so the reader docs here live in the Swagger and
DevTools guides and link to it. If #634 merges first the links resolve immediately; if this merges
first they're dead for one merge.

### DIFF
--- a/.changeset/route-flag-readers.md
+++ b/.changeset/route-flag-readers.md
@@ -1,0 +1,21 @@
+---
+'@forinda/kickjs': minor
+'@forinda/kickjs-swagger': minor
+'@forinda/kickjs-devtools': minor
+---
+
+Route flags reach the OpenAPI spec and the DevTools dashboard (phase 4 of `route-flags-design.md`).
+
+**`getRouteFlags(controllerClass, handlerName)`** resolves a route's flags from the controller — the same method-over-class result `ctx.route.flags` carries at request time, for consumers that see a controller and a method name rather than a live request: an adapter's `onRouteMount`, spec generation, tooling.
+
+**Swagger gains `publicFlag`.** Name the flag your project uses for public endpoints and the spec reads the same declaration the runtime does, instead of asking for a second annotation that can drift from it:
+
+```ts
+export const Public = defineRouteFlag('auth.public')
+
+SwaggerAdapter({ bearerAuth: true, publicFlag: 'auth.public' })
+```
+
+The name is configuration rather than a constant, because the framework deliberately names no flags — one project's `auth.public` is another's `public` or `security.none`. A list accepts several. It sits after `@ApiPublic` and `securityResolver` in the resolution order and before the `@ApiSecurity` / `@ApiBearerAuth` decorators, so an explicit resolver still wins while a flag still overrides class-level security.
+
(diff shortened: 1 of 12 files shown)


forinda:
Cross-referencing: I said earlier that this PR and #636 would conflict in `docs/guide/route-flags.md`. That was wrong — the file sets are disjoint (this PR touches `swagger.md` + `devtools.md`; #636 touches `route-flags.md` + `rate-limiting.md`), so there is no git conflict in either merge order.

The real gap was content: this PR's links point at a Consumers section in `route-flags.md` that never mentioned the readers, because this branch predates that file. #636 now carries the section documenting `publicFlag`, `getRouteFlags` and the DevTools column (commit bf57ce2), so the links resolve once both land.

Merge order still doesn't matter for conflicts — only that #634 lands before the links are live.

PR #637, opened 2026-09-03
feat: KickRouteFlags registry, its typegen, and runtime-aware engine peers

forinda:
Three commits pushed to #634's branch after it merged, so they never landed. Rebased onto `main`,
unchanged.

## 1. `KickRouteFlags` — flags get the type safety contributors have

Flag names were plain strings, which the design doc's own risk section called out: a typo is a
flag that silently never matches, and `flags.get()` returned `unknown`.

```ts
declare module '@forinda/kickjs' {
  interface KickRouteFlags {
    'auth.public': true
    'rate.limit': { rpm: number }
  }
}
```

- `defineRouteFlag('auth.pubic')` → compile error, with TypeScript's *"Did you mean"* suggestion
- `defineRouteFlag('rate.limit')` infers `RouteFlag<{ rpm: number }>` from the registry
- `flags.get('rate.limit')` is typed; `has()` takes only declared names
- `skipWhen` / `onlyWhen` / `exemptWhen` narrow in **every** consumer

Opt-in by construction: empty registry → everything falls back to `string` / `unknown`, the same
`[Known] extends [never]` trick `ContextMetaKey` uses. Projects that declare nothing compile
unchanged.

`defineRouteFlag` is overloaded rather than re-parameterised, so the documented
`defineRouteFlag<{ rpm: number }>('rate.limit')` keeps working next to the inferred form — the
first draft made the value generic positional and quietly broke it, which the type tests caught.

## 2. `kick typegen` generates that registry

Declaring the flag becomes the only step — no hand-written `declare module`:

```ts
// src/flags.ts — all you write
export const Public = defineRouteFlag('auth.public')
export const Limit = defineRouteFlag<{ rpm: number }>('rate.limit')
```

The new `kick/route-flags` plugin emits `.kickjs/types/kick__route-flags.d.ts`, refreshed on every
`kick dev` save.

**It is the cheapest plugin in the set, and gets more for it.** A flag is a positional string
literal with no `dependsOn` graph and no per-route ordering, so discovery is one AST branch and the
output is a flat map — where the context-key registry must reason about the contributor pipeline
first. It also captures the *value type*, so `flags.get('rate.limit')` is typed rather than merely
known to exist, which `ContextKeys` cannot do for contributors.

## 3. `kick add` installs the engine peers you actually use

The catalog listed `express` as a static peer of `@forinda/kickjs`, so adding the framework to a
Fastify or h3 project installed Express. The engine is chosen at `bootstrap({ runtime })`;
`resolveAppRuntime()` already existed for `kick add upload` and the framework package just never
used it.

| Runtime | Installed with `@forinda/kickjs` |
| --- | --- |
| `express` | `express` |
| `fastify` | `fastify`, `@fastify/middie`, `serve-static` |
| `h3` | `h3`, `serve-static` |

A test asserts these match what `kick new` scaffolds — drift shows up as a missing peer at boot,
a poor place to learn about it.

Also `kick typegen --list`, the answer to "how do I know what I can disable", now prints the file
each plugin owns and a copy-pasteable `typegen: { disable: [...] }` snippet. Its code comment
claimed it printed the owning plugin; it printed neither owner nor output file.

## Verification

- `pnpm typecheck` clean, including the `tsconfig.typetests.json` guards
- Repo suite: **315 files, 3291 tests, all passing**
- Three changesets: `@forinda/kickjs` minor, `@forinda/kickjs-cli` minor + patch

## Summary by CodeRabbit

* **New Features**
  * Added opt-in type safety for route flags, including validated names and inferred value types.
  * Route-flag conditions now support negated names, with mixed-polarity lists rejected early.
  * Use `@Flag.off` or `@Public.off` to remove inherited flags while preserving `false` as a value.
  * `kick typegen` generates route-flag declarations and refreshes them during development.
  * `kick add` selects engine-specific peer packages for Express, Fastify, or h3.
  * `kick typegen --list` displays generated files, watched inputs, and plugin disable guidance.

* **Documentation**
  * Added guidance for typed route flags, negation, flag removal, runtime-specific dependencies, and enhanced type generation output.

### DIFF
--- a/.changeset/add-engine-peers.md
+++ b/.changeset/add-engine-peers.md
@@ -1,0 +1,19 @@
+---
+'@forinda/kickjs-cli': patch
+---
+
+`kick add` installs the engine peers your project actually uses.
+
+The catalog listed `express` as a static peer of `@forinda/kickjs`, so `kick add kickjs` pulled Express into a Fastify or h3 project. The HTTP engine is chosen at `bootstrap({ runtime })` — which engine package a project needs is a runtime question, not a fixed dependency.
+
+It now resolves from the project's runtime (the `runtime` field in `kick.config.ts`, else the engine already in `package.json`), matching what `kick new` scaffolds for the same engine:
+
+| Runtime   | Installed with `@forinda/kickjs`             |
+| --------- | -------------------------------------------- |
+| `express` | `express`                                    |
+| `fastify` | `fastify`, `@fastify/middie`, `serve-static` |
+| `h3`      | `h3`, `serve-static`                         |
+
+`kick add --list` names the resolved engine on the row, and the install prints a notice saying which peers it added and why. `kick add upload` already worked this way; this brings the framework package itself in line.
+
+Also: `kick typegen --list` now prints the file each plugin owns and a copy-pasteable `typegen: { disable: [...] }` snippet, since "which plugin writes this file" is the question you have when deciding to turn one off.
--- a/.changeset/add-engine-peers.md
+++ b/.changeset/add-engine-peers.md
@@ -14,6 +14,10 @@ It now resolves from the project's runtime (the `runtime` field in `kick.config.
 | `fastify` | `fastify`, `@fastify/middie`, `serve-static` |
 | `h3`      | `h3`, `serve-static`                         |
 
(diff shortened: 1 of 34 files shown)


forinda:
@coderabbitai full review

PR #653, opened 2026-09-04
feat(kickjs): module-level route flags, and health probes that accept them

forinda:
## The gap

The built-in health probes mount as an ordinary module *inside* the middleware chain — deliberately, so a framework route never quietly bypasses app auth. But that left no good way to exempt them. A route flag can only be declared by a decorator, and `HealthController` is the framework's class, so the only option was exempting `/health/live` and `/health/ready` by pathname — exactly the brittle thing route flags were introduced to replace.

`health: false` isn't an escape either: it also skips the `HEALTH_PROBE` binding, so a replacement module can't reuse the draining/adapter-check logic.

## The module registration site

`ModuleRoutes.flags` applies flags to every route a mount produces — the site a decorator structurally cannot reach, because a module mounts controllers it may not own. This is the third level of the precedence chain the [roadmap already named](https://kickjs.app/guide/route-flags):

```ts
routes: () => ({ path: '/webhooks', controller: WebhooksController, flags: ['auth.public'] })
```

- **Precedence: method > class > mount.** A class declaration of the same flag wins, and `@Flag.off` on a method still removes an inherited one.
- **Two input forms.** A list stores each flag as `true` (`['auth.public']`); the record form carries values (`{ 'rate.limit': { rpm: 10 } }`). Both narrow against `KickRouteFlags`.
- **`!name` is rejected at boot.** A declaration has no negative form — remove the flag instead of declaring `'!x'`, which could never be tested for.

### `getRouteFlags` parity

Mount flags have nowhere on the class to live, so `getRouteFlags(controllerClass, handlerName)` — the resolver `SwaggerAdapter({ publicFlag })` and DevTools' route listing both use — would not have seen them. That would mean a spec marking a route public while the runtime protects it. Mount declarations are recorded against the controller (a `WeakMap`, replaced not appended, so HMR doesn't grow it) and folded in at the lowest precedence, so all three readers give the same answer.

## Health

```ts
bootstrap({ health: { flags: ['auth.public'] } })
```

That's the whole wiring. The framework still names no flags — `auth.public` is yours — and every consumer already reading it (an auth contributor's `skipWhen`, a guard's `exemptWhen`, `SwaggerAdapter({ publicFlag })`) now covers both probes. `health: true` / `health: false` behave exactly as before.

## Tests

`packages/kickjs/__tests__/route-flags-mount.test.ts` — 9 cases:

- mount flags reach a controller carrying no flag decorator
- `@Flag.off` on a method removes one
- a class declaration of the same flag wins
- the record form carries values
- `getRouteFlags()` agrees with the runtime, including the `.off` case
- `'!name'` throws at boot
- the probes are guarded by default, exempt with `health: { flags }`, and `health: false` still 404s

The health group uses an app-wide **contributor** rather than a global middleware, because global middleware runs pre-match and has no `ctx.route` — the same distinction the docs draw.

933 passed in `packages/kickjs`; swagger (62), devtools (94) and testing (229) green; lint and format clean.

## Docs

`guide/route-flags.md` (mount site + a health section), `guide/lifecycle.md` (the `/health` tip now shows the flag instead of "exempt the path"), `api/core.md`, `api/http.md`.

https://claude.ai/code/session_01TVCyhU9ghEntXjQwzXvjhE

### DIFF
--- a/.changeset/olive-cameras-heal.md
+++ b/.changeset/olive-cameras-heal.md
@@ -1,0 +1,28 @@
+---
+'@forinda/kickjs': minor
+---
+
+Route flags can now be declared at the module mount, and the built-in health probes accept them.
+
+`ModuleRoutes.flags` applies flags to every route a mount produces — the site a
+decorator cannot reach, because a module mounts controllers it may not own.
+Precedence is method > class > mount, so a class or method declaration still
+wins and `@Flag.off` still removes an inherited flag.
+
+```ts
+routes: () => ({ path: '/webhooks', controller: WebhooksController, flags: ['auth.public'] })
+```
+
+The health module is the first consumer. Its two probes sit inside the
+middleware chain, so app-wide auth applies to them, and until now the only way
+out was exempting `/health/live` and `/health/ready` by pathname:
+
+```ts
+bootstrap({ health: { flags: ['auth.public'] } })
+```
+
+The framework still names no flags — the name is yours, and every consumer that
+already reads it (`skipWhen`, `exemptWhen`, `SwaggerAdapter({ publicFlag })`)
+covers the probes with no further wiring. `getRouteFlags()` reports mount flags
+too, so an OpenAPI spec or DevTools listing cannot disagree with what the
+runtime resolved.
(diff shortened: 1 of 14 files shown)


forinda:
@coderabbitai full review

PR #660, opened 2026-09-04
fix(db): correct serial detection and array element types on introspect (#649, #648)

### LINKED ISSUE #648: db introspect: array columns become text, losing the element type
An array column is rendered as `text`, with the real type left in a comment:

```ts
export const payroll_payrolls = table('payroll_payrolls', {
  deductions: text(/* TODO: varchar[] */).notNull(),
})
export const exam_class_students = table('exam_class_students', {
  students: text(/* TODO: int4[] */),
})
```

Rebuilding from that schema gives `text` where the database has `varchar[]` and `int4[]` — a column that silently accepts a different shape of data.

### Note

As with the enum case, the type it could not build is already in the comment, so the information is read and then dropped. `customType` renders it correctly today:

```ts
const varchar_array = customType<unknown[]>({ dataType: () => 'varchar[]' })
// renders: "deductions" varchar[] NOT NULL
```

so emitting that instead of `text` would close the gap without needing a new builder — though a first-class `array(varchar())` would be better, since `customType<unknown[]>` gives up the element type on the TypeScript side.

Two columns in a 242-table schema, so low volume — but it fails quietly, which makes it worse than the ones that error.

---
`@forinda/kickjs-db@7.2.1`, Postgres 16.

### LINKED ISSUE #649: db introspect: a column defaulting from a standalone sequence is rendered serial(), changing its nullability
`introspect` treats any `nextval(...)` default as `serial()`. That is wrong when the sequence is not owned by the column.

### Reproduction

```sql
-- live
organizations.code  integer  NULL  DEFAULT nextval('code_seq'::regclass)
```

`code_seq` is a standalone sequence — no column owns it (`pg_depend` has no `deptype = 'a'` row for it), and it is shared rather than generated for this column.

`introspect` renders:

```ts
code: serial(),
```

which emits `"code" serial NOT NULL` — so the rebuilt column is **NOT NULL** where the original is nullable, and it gets a private `organizations_code_seq` instead of using the shared one.

### Distinguishing the two

The check is whether the sequence is owned:

```sql
select s.relname
  from pg_class s
  left join pg_depend d
    on d.objid = s.oid and d.deptype = 'a' and d.classid = 'pg_class'::regclass
  join pg_namespace n on n.oid = s.relnamespace
 where s.relkind = 'S' and n.nspname = 'public' and d.refobjid is null;
```

Note that the sequence *name* is not the signal. Five other columns in the same schema default from sequences named after an older table — `academic_departments.id` uses `departments_id_seq`, `exam_categories.id` uses `categories_id_seq` — and those are genuinely owned serials that should stay `serial()`. Only ownership distinguishes them.

### Expected

- an owned sequence → `serial()` / `bigSerial()`, whatever its name;
- an unowned one → the underlying integer type with the default preserved, and the sequence itself emitted as `CREATE SEQUENCE` (there is currently no way to declare a standalone sequence in the DSL either, which is the related gap).

---
`@forinda/kickjs-db@7.2.1`, Postgres 16.

forinda:
Closes #649. Closes #648.

Both are introspect changing what a column *means*, rather than merely losing detail.

## #649 — a standalone sequence read as `serial`

`isSerialColumn` keyed on the default alone:

```ts
r.column_default.startsWith('nextval(') && (udt is int2 | int4 | int8)
```

But `CREATE SEQUENCE s` followed by `ALTER TABLE t ALTER c SET DEFAULT nextval('s')` produces exactly that default while the sequence is standalone — shared across tables, or deliberately independent of the column's lifetime. Reading those as `serial` was wrong twice:

- the default is collapsed to `null` for a serial, so **the link to the sequence disappeared**
- `serial` implies NOT NULL, so a **nullable column came back non-nullable**

Two extra conditions settle it. `pg_get_serial_sequence` returns a name only for a sequence the column *owns* — what `serial` creates, and what `DROP TABLE` takes with it. And the column must be NOT NULL: a serial whose NOT NULL was dropped is no longer expressible as `serial`, and re-imposing the constraint on the next migration would reject the rows that made someone drop it in the first place.

## #648 — arrays lost their element type, in two places

`text[]` and `integer[]` both rendered as `text(/* TODO */)`. Two independent causes, which is why fixing either alone wouldn't have shown up:

1. **introspect** returned PG's *internal* element name. `data_type` is just `'ARRAY'` and the element lives in `udt_name` as `_int4`; stripping the underscore gave `int4[]`, `bool[]`, `bpchar[]` — none of which the DSL uses.
2. **render** had no array branch at all, so anything ending in `[]` missed every helper and hit the fallback.

Element names now map to the DSL surface, and `pickColumnHelper` splits the `[]` off so the column renders as `integer().array()`. `.array()` already existed on the column builder — nothing new in the DSL. An unmapped element type keeps its array-ness: `text(/* TODO: citext */).array()`.

Note that array columns come back unparameterised (`varchar[]`, not `varchar(255)[]`) — `information_schema` doesn't report `character_maximum_length` for arrays.

## Tests

The serial distinction only exists in a live database, so those go through testcontainers in `__tests__/integration/introspect-pg.test.ts`. One table, three `nextval()` columns, only one of them a serial:

- `id serial` → `serial`, default collapsed
- `ticket_no integer NOT NULL DEFAULT nextval('shared_counter')` → `integer`, default kept
- `maybe_no integer DEFAULT nextval('shared_counter')` → `integer`, **nullable**, default kept
- and separately, a `serial` with its NOT NULL dropped → `integer`, default kept

Plus `text[]` / `integer[]` against real Postgres — which is how I caught the `int4[]` half; the unit test alone would have passed with introspect still broken.

Five render cases: element helper + `.array()`, the import line carrying the element helper rather than `text`, parameterised element types, `.array()` ordering ahead of the nullability chain, and the unmapped-element fallback.

605 passing in `packages/db`, lint and format clean. Patch changeset.

https://claude.ai/code/session_01TVCyhU9ghEntXjQwzXvjhE

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL introspection now distinguishes standalone sequences from true serial columns.
  * Nullable columns with defaults are no longer incorrectly marked as `NOT NULL`.
  * PostgreSQL array column types now retain their element types and array behavior when rendered.
  * Array columns with mapped, parameterized, or unmapped element types are represented correctly.


forinda:
@coderabbitai full review

forinda:
@coderabbitai full review

Issue #672, opened 2026-09-06
Move the rules KickJS already has from runtime to build time


forinda:
# Move the rules KickJS already has from runtime to build time

## Context

Feedback from porting a large Express 4 + Sequelize app onto KickJS 8.3.1:
54 legacy apps, ~593 routes, 241 tables, database-per-tenant. Currently
~200 routes ported across 12 modules, 281 integration tests against real
Postgres.

The framework has held up well under that — the DI model, context
contributors and typegen are the reason the port is tractable. Almost every
hard problem has been the legacy app's, not KickJS's.

What follows is the one class of problem that has recurred: **rules the
framework already defines, but can only enforce once a request is in flight.**
Each has cost real debugging time, and each is discoverable statically.

---

## 1. A module cannot declare that it needs another module

`dependsOn` exists for plugins and adapters, not modules. So a module that
resolves a service out of another module compiles, boots, mounts its routes,
and then answers 500 on the first request that reaches one.

In this app:

- `ProcurementModule` → `StoresModule` (stocking a delivery goes through
  `AddStockUseCase`) and `FinanceModule` (paying a voucher posts a transaction)
- `StudentFinanceModule` → `FinanceModule` (fee structures resolve financial
  years; reception posts through the engine)

12 cross-module dependencies in total. All that held them together was a
sentence in each module's docstring:

```ts
/**
 * **Mount `StoresModule` and `FinanceModule` alongside this one.** … A module
 * cannot declare a mount dependency (`dependsOn` is plugin- and adapter-only),
 * so without them `POST /procurement/deliveries/:id/stock` and
 * `POST /procurement/payment-vouchers/:id/pay` fail to resolve.
 */
```

**Ask**: `dependsOn` on `defineModule`, validated at bootstrap, failing the way
a contributor's `dependsOn` already does (`MissingContributorError` at boot,
not first request). Modules already topo-sort for registration; this is the
same graph.

**Workaround in use** — a 90-line script wired into `kick ci:check` that walks
`from '@/modules/<other>/'` imports and asserts anything a mounted module
imports from is also mounted:

```
$ kick check:module-deps
  module deps: 12 cross-module dependencies, all mounted

# with .mount(FinanceModule()) removed:
  - ProcurementModule imports from src/modules/finance, but FinanceModule is
    not mounted in src/modules/index.ts.
  - StudentFinanceModule imports from src/modules/finance, but FinanceModule is
    not mounted in src/modules/index.ts.
  exit=1
```

[...]

PR #673, opened 2026-09-06
feat(cli): teach kick explain the framework's own DI errors

forinda:
## Summary

`kick explain` now recognizes the four errors KickJS itself produces. Split out of #672 (item 2b) — the cheapest item in that list: data only, no framework change.

The registry worked (`kick explain "@Module is not a function"` matches `module-decorator-not-found`) but held eight matchers, none keyed on a string the framework throws. All four of these returned "No known-issue matched":

| Error | Source |
| --- | --- |
| `Cannot inject REQUEST-scoped "…" into SINGLETON "…"` | `core/container.ts:732` |
| `Circular dependency detected: A -> B -> A` | `core/container.ts:536` |
| `No provider for <token>` (KICK001) | `core/kick-errors.ts` |
| every route 404s in dev | not thrown at all — see below |

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] CI / Tooling

## Changes

- Four new entries in `packages/cli/src/explain/known-issues.ts`, registered in `KNOWN_ISSUES`.
- 8 new tests in `packages/cli/__tests__/explain.test.ts`, including ranking against `module-not-registered` in both directions.
- Patch changeset on `@forinda/kickjs-cli`.

### The 404 one is not an error string

`packages/vite/src/dev-server.ts:148` reads `mod.app`, finds no `.handle`, and falls through to Vite's own 404. Nothing is logged. Every route 404s, including ones that obviously exist, and only in dev — `kick start` runs the entry directly and never needs the export.

The existing `module-not-registered` entry matched that 404 at confidence 50, which is the wrong diagnosis: it sends someone to audit a modules array that is already correct.

Getting the boundary right took two passes. The first version fired on any dev-server 404, which CodeRabbit correctly flagged: `vite dev: GET /typo returns 404` would have been diagnosed as a missing `app` export. A 404 now needs corroboration from one of two signals — the user naming the export, or the giveaway that it is not one route failing but all of them. `'no routes'` is deliberately excluded from that second list, since `'no route'` is already a 404 signal and would make every single-route 404 read as a total failure.

### One thing the runtime error gets wrong

The scope message ends with:

> Singletons outlive requests. Use TRANSIENT or REQUEST scope for the parent.

Good advice when the parent is a `@Service` / `@Repository` / `@Component` / `@Injectable` — each takes `{ scope }`. It is not available for a controller: `@Controller()` takes no options at all and always registers SINGLETON (`core/decorators.ts:178`), which is the exact shape reported in #672.

The fix there is `@Autowired`. Its getter re-resolves on every access for REQUEST and TRANSIENT dependencies and memoizes only SINGLETON ones (`core/container.ts:862-885`), so each request reads its own instance and the singleton never captures one. The diagnosis says this outright rather than repeating advice the reader cannot act on. The docs already point the same way — `docs/guide/dependency-injection.md:436` says "resolve it explicitly inside a method call rather than injecting it as a dependency", and the request-scope example uses `@Autowired` on a controller.

Wording that hint by parent kind at the throw site is worth its own change, since the container knows whether the target is a controller. Noted in #676.

## Related Issues

Part of #672. Related: #676.

## Checklist

- [x] `pnpm build` passes
- [x] `pnpm test` passes — `explain.test.ts` 31 passed; full CLI suite 816 passed / 76 files
- [x] `pnpm format:check` passes
- [ ] Docs updated — not applicable; every doc URL in the new entries points at an existing heading (`#scopes`, `#circular-dependency-detection`, `#registering-services`, `guide/project-structure`)
- [ ] New exports — not applicable

## Summary by CodeRabbit

* **New Features**
  * `kick explain` now recognizes additional dependency-injection issues, including scope mismatches, circular dependencies, and missing providers.
  * Added guidance for development-server 404 errors caused by missing `app` exports.
  * Explanations include diagnoses and recommended fixes, including controller-scope and `@Autowired` guidance.

* **Tests**
  * Added coverage for the new diagnostics, overlapping error precedence, development-server 404 scenarios, and production 404 handling.

### DIFF
--- a/.changeset/plain-pugs-explain.md
+++ b/.changeset/plain-pugs-explain.md
@@ -1,0 +1,7 @@
+---
+'@forinda/kickjs-cli': patch
+---
+
+`kick explain` now recognizes the framework's own DI errors: the REQUEST-into-SINGLETON scope violation, `Circular dependency detected`, `No provider for <token>` (KICK001), and the dev-server 404 that means the entry file exports no `app`.
+
+The scope entry notes what the runtime error cannot: `@Controller()` takes no `scope` option, so "use TRANSIENT or REQUEST scope for the parent" has no controller-shaped answer — `@Autowired` is the fix there.
--- a/packages/cli/__tests__/explain.test.ts
+++ b/packages/cli/__tests__/explain.test.ts
@@ -165,6 +165,57 @@ describe('known-issues registry', () => {
     expect(m?.diagnosis.id).not.toBe('test-env-leaked-from-dotenv')
   })
 
+  it('matches di-request-scope-into-singleton on the container error verbatim', () => {
+    const m = findBestMatch(
+      'Cannot inject REQUEST-scoped "TenantContext" into SINGLETON "ReceptionController". ' +
+        'Singletons outlive requests. Use TRANSIENT or REQUEST scope for the parent.',
+    )
+    expect(m).not.toBeNull()
+    expect(m!.diagnosis.id).toBe('di-request-scope-into-singleton')
+    expect(m!.confidence).toBe(100)
+  })
+
+  it('points a controller at @Autowired, since @Controller() takes no scope', () => {
+    // The framework's own message says "use TRANSIENT or REQUEST scope for the
+    // parent" — advice that does not exist for a controller. The diagnosis has
+    // to say so or it sends people looking for an option that isn't there.
+    const m = findBestMatch('Cannot inject REQUEST-scoped "X" into SINGLETON "YController"')
+    expect(m!.diagnosis.fix).toContain('@Autowired')
+    expect(m!.diagnosis.fix).toContain('@Controller() does NOT')
+  })
+
+  it('matches di-circular-dependency with the resolution chain', () => {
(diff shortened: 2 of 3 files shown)


forinda:
Fixed in 3934b7bc — the finding was right. `vite dev: GET /typo returns 404` scored 70 and outranked `module-not-registered`, diagnosing one missing route as a missing `app` export.

Took the correction rather than the proposed diff, though. Requiring `mentionsApp` outright would drop the case the entry is most useful for: someone who does not yet know the cause types "kick dev, every route 404s" and never says "export". So a 404 now needs one of two corroborating signals — the export named, or the giveaway that it is all routes failing rather than one.

`'no routes'` is deliberately absent from that second list: `'no route'` is already a 404 signal, so including it would make every single-route 404 read as a total failure and put the bug straight back.

Both regression cases added — the all-routes phrasing without the word "export" still matches, and the single-route Vite 404 falls back to `module-not-registered`. 31 tests in the file.

PR #703, opened 2026-09-14
fix(ws): remove upgrade listeners on shutdown so dev reloads keep WebSockets

forinda:
## Why

After a dev hot reload, every WebSocket connection answers **503** until `kick dev` is restarted.

On reload, `bootstrap()` calls `shutdown({ closeServer: false })` on the old app and starts a fresh one on the same HTTP server (`bootstrap.ts`). `WsAdapter.shutdown()` closed its `WebSocketServer` but never removed the `'upgrade'` listener it added in `afterStart`. Node runs upgrade listeners in registration order, so the stale listener ran first, matched `/ws/...`, and passed the socket to the closed server — `ws` rejects an upgrade on a closed server with 503 (`websocket-server.js`, `_state > RUNNING`) and destroys the socket before the new adapter's listener runs. Each save added another dead listener.

## What

- **`WsAdapter`** — the upgrade handler is held by reference; `shutdown()` removes it from the server.
- **`SocketIoAdapter`** — same class of bug. engine.io's `attach()` adds `upgrade` / `close` / `listening` listeners and swaps the server's own `request` listeners for a wrapper; `shutdown()` undid none of it, leaving a dead engine on the server and the app's request handler (Vite's, in dev) wrapped by it. `afterStart` now records the listeners before `attach()`, and `shutdown()` removes what `attach()` added and restores the request listeners it took.

## Tests

For each adapter, the reload sequence on one server: attach → shutdown → a fresh adapter attaches.
- `WsAdapter`: upgrade listener count returns to baseline; a client then connects (was 503).
- `SocketIoAdapter`: `request` / `upgrade` / `close` / `listening` counts return to baseline; the server's own request handler still answers; a client then connects and gets its echo.

Both fail against the previous code. Also checked through the real path (throwaway test, not committed): three `bootstrap()` reloads on a shared dev server — all three connections open, one upgrade listener remains.

ws suite 54 passed / 9 skipped (Redis/Centrifugo suites), typecheck (src + tests), oxlint, format.

Not included: a framework-level safety net in `Application.shutdown({ closeServer: false })` for third-party adapters that make the same mistake — possible follow-up.

Changeset: `@forinda/kickjs-ws` patch.

https://claude.ai/code/session_011wf3GNMSJUfME42a93d5j6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket and Socket.IO behavior during development hot reloads.
  * Prevented stale connections and handlers from causing failed WebSocket upgrades or unexpected 503 responses.
  * Ensured servers reused after reload continue serving regular requests and accept new WebSocket connections correctly.
  * Prevented listener buildup across repeated adapter restarts, improving development server stability.

* **Tests**
  * Added coverage for restarting adapters on shared HTTP servers and reconnecting clients successfully.


forinda:
@coderabbitai full review

